PR #6, opened 2021-09-22
h264 encoder patch

hiroshihorie:
Applies h264 encoder patch to prevent segmentation fault when downgrading simulcast.

Credit to the following patches
https://github.com/shiguredo-webrtc-build/webrtc-build/blob/master/patches/macos_simulcast.patch
https://github.com/shiguredo-webrtc-build/webrtc-build/blob/master/patches/macos_h264_encoder.patch


cloudwebrtc:
LGTM

PR #5, opened 2021-09-16
Release mic when category changes

hiroshihorie:
- Restart AudioUnit any time category changes between mic-using-category / non-mic-using category
- Rename symbols

PR #7, opened 2021-09-22
Change AVAudioSession defaults to iOS defaults

hiroshihorie:
Currently, the default category is `AVAudioSessionCategoryPlayAndRecord` which will automatically turn-on the microphone **even for listen/view-only cases**.

Instead of forcing developer to use `playAndRecord`, developer should be responsible to set the relevant category, which will result mic to turn on if `playAndRecord` or `record` is used. (PR https://github.com/webrtc-sdk/webrtc/pull/5)

This PR sets the defaults to iOS defaults.

https://developer.apple.com/documentation/avfaudio/avaudiosession/capturing_stereo_audio_from_built-in_microphones

### DIFF
--- a/sdk/objc/components/audio/RTCAudioSessionConfiguration.m
+++ b/sdk/objc/components/audio/RTCAudioSessionConfiguration.m
@@ -69,11 +69,11 @@ static RTC_OBJC_TYPE(RTCAudioSessionConfiguration) *gWebRTCConfiguration = nil;
     // By default, using this category implies that our app’s audio is
     // nonmixable, hence activating the session will interrupt any other
     // audio sessions which are also nonmixable.
-    _category = AVAudioSessionCategoryPlayAndRecord;
-    _categoryOptions = AVAudioSessionCategoryOptionAllowBluetooth;
+    _category = AVAudioSessionCategorySoloAmbient;
+    _categoryOptions = 0;
 
     // Specify mode for two-way voice communication (e.g. VoIP).
-    _mode = AVAudioSessionModeVoiceChat;
+    _mode = AVAudioSessionModeDefault;
 
     // Set the session's sample rate or the hardware sample rate.
     // It is essential that we use the same sample rate as stream format


davidzhao:
this looks good to me. question is how does this work for existing flutter-webrtc users? LiveKit iOS SDK already handles management of the audio session.

feels like flutter-webrtc should add this management layer before upgrading to this build. here's the logic that we use to [init audio session](https://github.com/livekit/client-sdk-ios/blob/e7af89dc223cc26826203a8dc2016941a120f88c/Sources/LiveKit/classes/Room.swift#L243) when the first audio track is received. and [here is when it's published](https://github.com/livekit/client-sdk-ios/blob/c85c6cadb50245fe9a6ccf32e13b66a72bc84301/Sources/LiveKit/classes/LocalParticipant.swift#L41)

@cloudwebrtc, do you want to add this to flutter-webrtc first?

hiroshihorie:
@davidzhao Thanks for pointing this out, this is my concern also.

cloudwebrtc:
hey, @davidzhao @hiroshihorie I think we need to find whether there is a [localStream](https://github.com/flutter-webrtc/flutter-webrtc/blob/master/common/darwin/Classes/FlutterWebRTCPlugin.h#L18) (including audio track) in the flutter-webrtc plugin to switch `category`, when there is no loalstream created, there is no need to use `playAndRecord`

davidzhao:
> These operations can be done internally by flutter-webrtc native, and users have no perception.

Yes! I agree this could be done transparently by flutter-webrtc.

I think only two actions are needed:
* when user creates a local audio track, set audio session to playAndRecord
* when a onTrack event is triggered, and it's an audio track, set audio session to play (if it's currently ambient)

hiroshihorie:
@cloudwebrtc @davidzhao Yes I agree it should be handled internally, but there should be an option for this to be off. So if the developer is managing the audio session themself it will cause conflict.

davidzhao:
if we only set audiosession if it's set to ambient, then it should be ok right?

if dev set it to something else, we would not override.

similarly with playAndRecord, we would only set it if it wasn't set

hiroshihorie:
Should be ok most of the time but I'm still thinking if it's ok for the uncommon cases.

Some `categoryOption`s  and `mode`s don't work with different categories, and switching category transparently might reset these settings.

I think it still might be safe to have an option to turn off this auto-management of audio session, since it's hard to understand 100% of developer's intention.

`soloAmbient` does not support background playback so it is possible that developer would like to change this to `playback` https://developer.apple.com/documentation/avfaudio/avaudiosession/category/1616509-playback

**Could be almost fully managed :**
No LOCAL AUDIO: Set one of `playback` (dev may want `soloAmbient`, `ambient` for some reason)
Has LOCAL AUDIO track: Set `record`
Has LOCAL AUDIO track and REMOTE AUDIO track: Set`playAndRecord`

Problem 1: How to know user's preferred playback mode (`playback`, `soloAmbient`, `ambient`)
Possible solution: Store the initial playback category and set it back to that.

Problem 2: How should we handle `categoryOption`s  and `mode`s ?

I'm not sure about `multiRoute` category. 😅

davidzhao:
> Some `categoryOption`s and `mode`s don't work with different categories, and switching category transparently might reset these settings.

You are right that it's possible they might want ambient. I think it's ok for users to actually tell us those use cases.. and we can add a toggle then to disable this? Just trying to avoid over-engineering this for the initial version. So proposal is:

1. set to `playback` (with appropriate mode) if current category is ambient AND a remote audio track is received
2. set to `playAndRecord` (with appropriate mode) if current category is *not* playAndRecord AND a local track is created.

wdyt?

hiroshihorie:
@davidzhao Sounds good! won't hurt to have an option to turn this off tho!

davidzhao (approved):
👍

PR #8, opened 2021-09-28
Sync audio session config

hiroshihorie:
### Fixes Issue
Mic doesn't turn on when there are no remote audio tracks.
This does fix the issue but I'm not satisfied with it, there might be more changes.

### Changes

1. Use AVAudioSession values when initializing `RTCAudioSessionConfiguration`.
2. Always restart AudioUnit when category changes.
for example, .playAndRecord -> .record didn't restart it before this change.

~~2. Update `RTCAudioSessionConfiguration.webRTCConfiguration` when category changes.
Internally, `configureWebRTCSession` is called and `RTCAudioSessionConfiguration.webRTCConfiguration` is referenced. The values must be in sync with AVAudioSession for expected behavior.~~

### DIFF
--- a/sdk/objc/components/audio/RTCAudioSession.mm
+++ b/sdk/objc/components/audio/RTCAudioSession.mm
@@ -103,8 +103,6 @@ NSString * const kRTCAudioSessionOutputVolumeSelector = @"outputVolume";
                   options:NSKeyValueObservingOptionNew | NSKeyValueObservingOptionOld
                   context:(__bridge void *)RTC_OBJC_TYPE(RTCAudioSession).class];
 
-    _isRecordingEnabled = [self sessionCategoryIsRecordingEnabled];
-
     _activeCategory = _session.category;
 
     RTCLog(@"RTC_OBJC_TYPE(RTCAudioSession) (%p): init.", self);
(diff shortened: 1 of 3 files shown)


sdk/objc/components/audio/RTCAudioSession.mm
```diff
@@ -494,9 +496,16 @@ - (void)handleRouteChangeNotification:(NSNotification *)notification {
     case AVAudioSessionRouteChangeReasonCategoryChange:
```
davidzhao, line 496: FYI, these events are only triggered after the AVAudioSession has been activated. Initially it remains inactive until WebRTC decides to start it.

sdk/objc/components/audio/RTCAudioSession.mm
```diff
@@ -494,9 +496,16 @@ - (void)handleRouteChangeNotification:(NSNotification *)notification {
...
-        BOOL newValue = [self sessionCategoryIsRecordingEnabled];
-        if (_isRecordingEnabled != newValue) {
-          _isRecordingEnabled = newValue;
+        // webRTCConfiguration is used by configureWebRTCSession
+        RTC_OBJC_TYPE(RTCAudioSessionConfiguration) *webRTCConfig = [RTC_OBJC_TYPE(RTCAudioSessionConfiguration) webRTCConfiguration];
+        webRTCConfig.category = _session.category;
+        webRTCConfig.categoryOptions = _session.categoryOptions;
+        webRTCConfig.mode = _session.mode;
+
+        if (![_session.category isEqualToString:_activeCategory]) {
```
davidzhao, line 505: should we get rid of `_isRecordingEnabled` then?

davidzhao (commented):
looks good! this is an improvement regardless.

hiroshihorie:
WebRTC developers intended changes in the following order:
1. Update [RTCAudioSessionConfiguration webRTCConfiguration]
2. RTCAudioSession.setConfiguration
3. This will internally call AVAudioSession

Instead we are
1. Changing AVAudioSession first
2. Then updating Update [RTCAudioSessionConfiguration webRTCConfiguration]

Which is not ideal.... but it shouldn't hurt.

sdk/objc/components/audio/RTCAudioSession.mm
hiroshihorie: Maybe we don't need it...

davidzhao (approved):
looks good!

PR #9, opened 2021-10-05
Add Apache-2.0 license and some note to README.md.

cloudwebrtc:
@voluntas @davidzhao @hiroshihorie I added a patch for the missing Apache-2.0 License. I have clarified my contribution. If you have any questions, please comment below.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,3 +29,11 @@ native API header files.
  * [Reporting bugs](docs/bug-reporting.md)
 
 [native-dev]: https://webrtc.googlesource.com/src/+/refs/heads/master/docs/native-code/index.md
+
+### Note
+
+This is a community maintenance project, not the official Google WebRTC repository. It is maintained by some contributors with WebRTC development experience, and the goal is to create a flexible binary release version.
+
+* The original code was cloned from [google webrtc](https://chromium.googlesource.com/external/webrtc.git) and released with [BSD license](/LICENSE).
+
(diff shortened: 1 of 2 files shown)


cloudwebrtc:
Perhaps we should also include an explanation in the binary package that part of the patched code uses the license in this link.
https://github.com/shiguredo-webrtc-build/webrtc-build#%E3%83%A9%E3%82%A4%E3%82%BB%E3%83%B3%E3%82%B9

cloudwebrtc:
@davidzhao I think you should know Apache-2.0 better than me :) (most of my projects use MIT), so I don't know the files that must be added and distributed with the binary. Do you have any suggestions?

README.md
```diff
@@ -29,3 +29,7 @@ native API header files.
  * [Reporting bugs](docs/bug-reporting.md)
 
 [native-dev]: https://webrtc.googlesource.com/src/+/refs/heads/master/docs/native-code/index.md
+
+### Note
```
cloudwebrtc, line 33: I am a bit confused whether we should make changes in the original readme file of webrtc, I don't want to modify it, but I still want to explain these contributions from the community through a readme.

NOTICE
```diff
@@ -0,0 +1,26 @@
+###################################################################################
+
+The following modifications follow Apache License 2.0 from shiguredo.
+
```
cloudwebrtc, line 4: So far, we have used three patches from Shiguredo, I think we will follow more Shiguredo patches.

voluntas:
> I think that in addition to declaring the source of these PRs in the patched repo, when releasing the binary, we still need to specify the part authorized by Apache-2.0 in the LICENSE file in the released binary file.

This recognition is correct.

README.md
voluntas: I think it would be better to create a file like README_webrtc_sdk.txt.

voluntas:
For clarity, we have prepared a NOTICE file. It is a good idea to explicitly post the copyright written here.

https://github.com/shiguredo-webrtc-build/webrtc-build/blob/master/NOTICE

README.md
davidzhao: I would vote for modifying the main README, first explaining the intent of the repo (i.e. provide mobile SDKs), and the changes we are applying to it.

Then include original Google readme at the bottom.

voluntas:
LGTM

PR #15, opened 2021-12-28
feat: support bypass voice processing for iOS.

cloudwebrtc:
Support bypass voice processing.
For example, singing, real-time collection of musical instruments, etc.

With the closure of AEC, NS, AGC, and high-pass filtering, we can use high-quality audio sources


davidzhao (approved):
looks good!

PR #18, opened 2022-02-09
Apply patches to m97

davidliu:
Note: PRs #11 and #12 were dropped as they were patches originally taken from the upstream, and are no longer needed.


davidzhao (approved):
Looks good! I've enabled individual commit merging for the repo. that might be a better way to move forward to keep them separate. wdyt?

davidliu:
What's individual commit merging, and how does that differ?

davidzhao:
it will keep all the commits in main individually, rather than all together. It might make it easier to cherry pick changes to future branches

davidliu:
Ah, instead of squashing, got it.

hiroshihorie:
Nice !! No more simulcast factory null checks for android 🙂

davidliu:
The null checks were originally from upstream so they're taken care of now.

PR #21, opened 2022-03-10
[Mac] Allow audio device selection

hiroshihorie:
### Goals

- [x]  Audio device list & selection at runtime.
- [x]  Detect and switch device when default device changes at runtime. (Connect AirPods etc)
- [ ] Feed audio sample buffers directly.

`bypassAudioProcessing` is _temporarily removed_, but I will make it available again.
-> Now available again.


modules/audio_device/mac/audio_device_mac.cc
```diff
@@ -867,12 +864,9 @@ int32_t AudioDeviceMac::RecordingDeviceName(uint16_t index,
   }
 
   memset(name, 0, kAdmMaxDeviceNameSize);
-
-  if (guid != NULL) {
```
davidzhao, line 852: can anything else call this function and pass in NULL for `RecordingDeviceName` ?

davidzhao (approved):
incredible work @hiroshihorie!

For the new files such as: RTCAudioDeviceModule.mm, are those from somewhere else in the webrtc tree?

modules/audio_device/mac/audio_device_mac.cc
hiroshihorie: `RecordingDeviceName` is supposed to populate `name` and `guid` of the device,
but populating `guid` was not implemented in their code.

So I implemented to return the `AudioDeviceID` which is not technically `guid` but 
it works for identifying the device.

hiroshihorie:
Yes there is a `webrtc::AudioDeviceModule` class.
https://chromium.googlesource.com/external/webrtc/stable/webrtc/+/15a8a6e7edb2c8df7f5b21e759ace213a4ab7292/modules/audio_device/include/audio_device.h#19

davidzhao:
> Yes there is a `webrtc::AudioDeviceModule` class. https://chromium.googlesource.com/external/webrtc/stable/webrtc/+/15a8a6e7edb2c8df7f5b21e759ace213a4ab7292/modules/audio_device/include/audio_device.h#19

But is the ObjC wrapper class from the tree? or is that a new file you authored?

hiroshihorie:
Yes there's a few new files I authored.

for `RTCAudioDeviceModule`, In the future,
I might want to add ability to create virtual devices.
`RTCAudioDeviceModule.addVirtualDevice(...)`

then user can call the existing (implemented in this PR) `RTCAudioDeviceModule.setInputDevice(...)`,
to switch to their virtual device.

This virtual device class will conform to a new `RTCVirtualAudioDevice` protocol (or base class) so  users can feed their own `CMSampleBuffer`s through here.

This is just a thought for design 🙂

I want to achieve unified design that will work for both iOS / Mac platforms, but the underlying audio system is pretty different for each platform. So I'm not sure if it's possible.

Another concern is mixing the audio, for example game streaming, user would like to speak while having game audio. `ReplayKit` delivers audio buffer separately for `app` and `mic`.
https://developer.apple.com/documentation/replaykit/rpsamplebuffertype
So it needs to get mixed somehow... still investigating ...

davidzhao:
Got it, for the new files, could you add a LiveKit copyright in it? This will help differentiate which files originated from upstream vs us.

davidzhao:
> This virtual device class will conform to a new `RTCVirtualAudioDevice` protocol (or base class) so users can feed their own `CMSampleBuffer`s through here.
> 
> This is just a thought for design 🙂
> 
> I want to achieve unified design that will work for both iOS / Mac platforms, but the underlying audio system is pretty different for each platform. So I'm not sure if it's possible.

YES! This is going to be awesome!!

hiroshihorie:
RTC's thread checks were failing so improved thread handling.
All access to ADM methods are from worker thread now.

`RTC_DCHECK_RUN_ON(&main_thread_checker_);` checks whether its called from the same thread that the ADM was created on (which must be created on worker thread).
https://github.com/webrtc-sdk/webrtc/blob/main/modules/audio_device/audio_device_buffer.cc#L135
https://github.com/webrtc-sdk/webrtc/blob/main/modules/audio_device/audio_device_buffer.cc#L149

hiroshihorie:
## Additional required logic

### Active device (lost) disconnected ✅
Current: Nothing happens (audio input/output stops)
Expected: Switch to `default` device and continue input/output.

### OS's default device changed ✅
Current: Audio input/output continues on the same device even if `default` device is changed.
Expected: if `default` device is switched by the OS, change input/output to the new `default` device.

hiroshihorie:
Above 2 logic have been implemented and seems to work as expected.

cloudwebrtc (approved):

hiroshihorie:
implemented `audioDeviceModule.setDevicesUpdatedHandler` to get notification when
audio input / output devices update ( or default changes )

cloudwebrtc:
I think we need to compile it into an `xframework` to make sure the API works correctly on macOS and doesn't break iOS

hiroshihorie:
I made some changes and iOS compiles

hiroshihorie:
Fixed bug where `AudioDeviceSink` gets reset since `RTCAudioDeviceModule` was being recreated on every access.

cloudwebrtc:
hey @hiroshihorie, It seems that changing the PR from target m97 to m104 introduces additional diffs, should we still target m97 after merging we get the minimal diff for the patch?

hiroshihorie:
> hey @hiroshihorie, It seems that changing the PR from target m97 to m104 introduces additional diffs, should we still target m97 after merging we get the minimal diff for the patch?

Yup, I changed the target which wasn't a good idea..

cloudwebrtc:
The code looks good.

davidzhao:
@hiroshihorie shall we merge this as well as into 104?

PR #23, opened 2022-05-05
[Windows] Fixed the audio capture issue when the built-in AEC enabled

cloudwebrtc:
Fixed the issue when the built-in AEC is valid under Windows, the mic cannot be captured when StartRecording is called before StartPlayout.

Trigger conditions for this bug:
1, windows and supports built-in AEC enabled
2. StartRecording is called before StartPlayOut, such as SendOnly mode, or AudioSendStream is enabled before AudioRecvStream

The raised issue, unable to capture mic and no audio stream sent

### DIFF
--- a/audio/audio_state.cc
+++ b/audio/audio_state.cc
@@ -104,6 +104,14 @@ void AudioState::AddSendingStream(webrtc::AudioSendStream* stream,
   if (!adm->Recording()) {
     if (adm->InitRecording() == 0) {
       if (recording_enabled_) {
+#if defined(WEBRTC_WIN)
+        if (adm->BuiltInAECIsAvailable() && !adm->Playing()) {
+          if (!adm->PlayoutIsInitialized()) {
+            adm->InitPlayout();
+          }
+          adm->StartPlayout();
+        }
+#endif
         adm->StartRecording();
       }
     } else {
--- a/modules/audio_device/win/audio_device_core_win.cc
+++ b/modules/audio_device/win/audio_device_core_win.cc
@@ -2356,7 +2356,7 @@ int32_t AudioDeviceWindowsCore::StartRecording() {
       }
     }
 
-    RTC_DCHECK(_hRecThread);
+    RTC_DCHECK(_hRecThread == NULL);
     _hRecThread = CreateThread(NULL, 0, lpStartAddress, this, 0, NULL);
     if (_hRecThread == NULL) {
       RTC_LOG(LS_ERROR) << "failed to create the recording thread";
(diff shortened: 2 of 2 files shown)


davidliu:
Hey, can you do this against the m97_release branch? Main is the old m93 release.

modules/audio_device/win/audio_device_core_win.cc
```diff
@@ -2356,7 +2356,7 @@ int32_t AudioDeviceWindowsCore::StartRecording() {
       }
     }
 
-    RTC_DCHECK(_hRecThread);
+    RTC_DCHECK(_hRecThread == NULL);
```
davidliu, line 2359: Looks like this is fixed in m97.

modules/audio_device/win/audio_device_core_win.cc
```diff
@@ -2493,7 +2493,7 @@ int32_t AudioDeviceWindowsCore::StartPlayout() {
     MutexLock lockScoped(&mutex_);
 
     // Create thread which will drive the rendering.
-    RTC_DCHECK(_hPlayThread);
+    RTC_DCHECK(_hPlayThread == NULL);
```
davidliu, line 2496: Looks like this is fixed in m97.

audio/audio_state.cc
```diff
@@ -104,6 +104,14 @@ void AudioState::AddSendingStream(webrtc::AudioSendStream* stream,
   if (!adm->Recording()) {
     if (adm->InitRecording() == 0) {
       if (recording_enabled_) {
+#if defined(WEBRTC_WIN)
+        if (adm->BuiltInAECIsAvailable() && !adm->Playing()) {
+          if (!adm->PlayoutIsInitialized()) {
+            adm->InitPlayout();
+          }
+          adm->StartPlayout();
```
davidliu, line 112: Could you write a comment explaining why this is needed? Seems weird to require playout for recording.

cloudwebrtc:
sure, I forgot about this. xD

audio/audio_state.cc
cloudwebrtc: The problem is caused here https://github.com/webrtc-sdk/webrtc/blob/main/modules/audio_device/win/audio_device_core_win.cc#L2353. If the built-in AEC is enabled, opening the Recording alone will fail to create hRecThread, resulting in no sound captured by the send stream.

But if StartPlayOut is called early, it works both ways.

cloudwebrtc: I think the code here is a bit strange, with the built-in AEC enabled, recording/playout cannot be started independently. ,

cloudwebrtc:
new PR here. https://github.com/webrtc-sdk/webrtc/pull/29

PR #25, opened 2022-05-09
Start/Stop receiving stream method for VideoTrack

davidliu:
Used to mitigate decoder limitations (android crashes if there are too many decoders, iOS freezes up).


hiroshihorie:
Wow, would like to try this out !

davidzhao (commented):
Incredible work! This is a huge deal 🔥

PR #34, opened 2022-07-18
m104 update

davidliu:
Having issues making the iOS framework (not lib related, generating licenses is giving me errors), but builds fine for Android. @cloudwebrtc can you try building and see if you run into any issues?


cloudwebrtc:
> Having issues making the iOS framework (not lib related, generating licenses is giving me errors), but builds fine for Android. @cloudwebrtc can you try building and see if you run into any issues?

sure, I'll try to build this branch later.

davidliu:
Nevermind, was using the wrong build command, iOS builds fine, I'll run through some sanity tests on it in a bit.

davidzhao (approved):
excellent! Is the idea of having a single PR with our patches to make it easier for future upgrades?

davidliu:
> excellent! Is the idea of having a single PR with our patches to make it easier for future upgrades?

I think we'll stick with the rebase and merge for the time being. There were some merge conflicts, and it's easier to view how to resolve those in the context of each commit.

ycherniavskyi:
Also, it would be great if someone dropped several lines in [wiki](https://github.com/webrtc-sdk/webrtc/wiki) about this project git flow because from the side, it is not obvious at all.

cloudwebrtc (approved):

hiroshihorie (approved):

davidliu:
Hmm, I'm thinking we should update the default branch every time we pull from upstream, so that people can see the latest changes when viewing the repo (main is on m93). Alternative is to save off the old versions in the separate branches and rewrite main each time? Thoughts?

davidzhao:
> 

agreed..changing the primary branch each time to the latest seem to be the most straight-forward option

ycherniavskyi:
I fully agree that the primary branch must point to the actually used version in `flutter_webrtc`. I am exactly one of those that look at the `main` branch and can't grasp why plugging works another way than code tell 🤪. And only in several hours, I understood that the plugin is on `m97_release` 😅.

PR #39, opened 2022-08-23
Expose E2EE interfaces

hiroshihorie:


sdk/objc/api/peerconnection/RTCFrameEncryptorInterface.h
```diff
@@ -0,0 +1,58 @@
...
+  // Attempts to encrypt the provided frame. You may assume the encrypted_frame
+  // will match the size returned by GetMaxCiphertextByteSize for a give frame.
+  // You may assume that the frames will arrive in order if SRTP is enabled.
+  // The ssrc will simply identify which stream the frame is travelling on. You
+  // must set bytes_written to the number of bytes you wrote in the
+  // encrypted_frame. 0 must be returned if successful all other numbers can be
+  // selected by the implementer to represent error codes.
+  // virtual int Encrypt(cricket::MediaType media_type,
+  //                     uint32_t ssrc,
+  //                     rtc::ArrayView<const uint8_t> additional_data,
```
hiroshihorie, line 35: @cloudwebrtc Any idea what the ObjC equivalent should be for `rtc::ArrayView<const uint8_t>` ?

cloudwebrtc: I think is `NSData`.

hiroshihorie:
@cloudwebrtc Looks like we will be exposing another interface, should I close this one ?

cloudwebrtc:
@hiroshihorie FrameEncryptorInterface does not provide the type of video frame (key, delta), and timestamp (for generating iv), so use https://github.com/webrtc-sdk/webrtc/blob/m104_release/api/frame_transformer_interface.h instead

cloudwebrtc:
I created FrameCryptor in this branch https://github.com/webrtc-sdk/webrtc/pull/54/files, it can interop with js e2ee (aes-gcm), and it is WIP currently.  
I believe it can be wrapped inside objc, and eventually, you can expose an Objc method in RTCRtpSender.h, such as EnableFrameCrypto(enum Algorithm, keyManager);

PR #50, opened 2022-11-11
fix: Restarting the playout without audio send streams makes the mic indicator off.

cloudwebrtc:
fix: Restarting the playout without audio send streams makes the mic indicator off.

### DIFF
--- a/audio/audio_state.cc
+++ b/audio/audio_state.cc
@@ -124,6 +124,17 @@ void AudioState::RemoveSendingStream(webrtc::AudioSendStream* stream) {
   UpdateAudioTransportWithSendingStreams();
   if (sending_streams_.empty()) {
     config_.audio_device_module->StopRecording();
+#if defined(WEBRTC_IOS)
+    auto* adm = config_.audio_device_module.get();
+    if (adm->Playing()) {
+      adm->StopPlayout();
+      if (adm->InitPlayout() == 0) {
+        if (playout_enabled_) {
+          adm->StartPlayout();
+        }
+      }
+    }
+#endif
   }
 }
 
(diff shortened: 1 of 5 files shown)


audio/audio_state.cc
```diff
@@ -124,6 +124,17 @@ void AudioState::RemoveSendingStream(webrtc::AudioSendStream* stream) {
   UpdateAudioTransportWithSendingStreams();
   if (sending_streams_.empty()) {
     config_.audio_device_module->StopRecording();
+#if defined(WEBRTC_IOS)
+    auto* adm = config_.audio_device_module.get();
+    if (adm->Playing()) {
+      adm->StopPlayout();
+      if (adm->InitPlayout() == 0) {
```
davidzhao, line 131: what does this mean? we are starting playout after stopping it?

davidzhao (approved):
this is a great fix! would help resolve mic indicators across our other platforms too.

mind adding a bit of comment there? both to indicate we added it but also a bit about how it works.

audio/audio_state.cc
cloudwebrtc: Through repeated tests, when the mode is changed to playback, the RTCAudioSession must be reset to stop the capture of the microphone.

cloudwebrtc: I'll add a commen describing the intent of this PR.

davidzhao: got it, so it'd have to be restarted after Playout is stopped?

cloudwebrtc: If playout is stopped, we should not need to restart it, only need to deal with `adm->StopRecording()  && adm->Playing()`

It means that after the last audio send stream is closed, 1,stop recording, 2, and then reset to release the mic capture state.

cloudwebrtc:
Replaced with a more reasonable PR. https://github.com/webrtc-sdk/webrtc/pull/52

PR #55, opened 2022-11-30
Stop recording on mute (turn off mic indicator)

hiroshihorie:
Now, if all audio tracks are muted(disabled), recording will stop and mic indicator will turn off.
Unmuting(enabling) any audio track will restart audio recording.

This is possible because of https://github.com/webrtc-sdk/webrtc/pull/52
Works(tested) with iOS / macOS.

macOS:
https://user-images.githubusercontent.com/548776/204782648-db38668f-21d6-46ac-876a-728376b716f2.mov


cloudwebrtc (approved):
lgtm

PR #66, opened 2023-03-22
feat: key ratchet/derive.

cloudwebrtc:


api/crypto/frame_crypto_transformer.cc
```diff
@@ -470,35 +517,88 @@ void FrameCryptorTransformer::decryptFrame(
...
-      if (observer_)
-        observer_->OnFrameCryptionError(participant_id_, last_dec_error_);
+    if (key_handler->options().ratchet_window_size > 0) {
+      while (ratchet_count < key_handler->options().ratchet_window_size) {
+        ratchet_count++;
+
+        RTC_LOG(LS_INFO) << "ratcheting key attempt " << ratchet_count << " of "
+                         << key_handler->options().ratchet_window_size;
+
+        key_handler->RatchetKey(key_index);
```
lukasIO: what I did in the JS implementation now is that 
`RatchetKey` returns the key directly and there's a parameter to specifiy whether or not to automatically set the key after ratcheting. 
This allows us to say `give me the next ratcheted key, but only set it on the key manager as the correct one, if we know that decryption was successful`.

cloudwebrtc: @lukasIO  updated https://github.com/webrtc-sdk/webrtc/pull/66/commits/faa419cee00cdab7780ee14532f531d0421ce167

and same changes for flutter https://github.com/livekit/client-sdk-flutter/pull/226/commits/7cad04427386f89b9ac589704b773647786b28e7

lukasIO (approved):
awesome stuff! 🚀

PR #70, opened 2023-05-18
Improve H264 encoder

hiroshihorie:
* Improve bitrate
* Turn on Apple's optimizations when available
* maxQP for screen share mode
* Some code clean up

### DIFF
--- a/sdk/objc/components/video_codec/RTCVideoEncoderH264.mm
+++ b/sdk/objc/components/video_codec/RTCVideoEncoderH264.mm
@@ -54,7 +54,7 @@ namespace {  // anonymous namespace
 // The ratio between kVTCompressionPropertyKey_DataRateLimits and
 // kVTCompressionPropertyKey_AverageBitRate. The data rate limit is set higher
 // than the average bit rate to avoid undershooting the target.
-const float kLimitToAverageBitRateFactor = 1.5f;
+const float kLimitToAverageBitRateFactor = 5.0f;
 // These thresholds deviate from the default h264 QP thresholds, as they
 // have been found to work better on devices that support VideoToolbox
 const int kLowH264QpThreshold = 28;
@@ -317,6 +317,7 @@ NSUInteger GetMaxSampleRate(const webrtc::H264ProfileLevelId &profile_level_id) 
   RTC_OBJC_TYPE(RTCVideoCodecInfo) * _codecInfo;
   std::unique_ptr<webrtc::BitrateAdjuster> _bitrateAdjuster;
   uint32_t _targetBitrateBps;
+  uint32_t _targetFrameRate;
   uint32_t _encoderBitrateBps;
   uint32_t _encoderFrameRate;
   uint32_t _maxAllowedFrameRate;
@@ -331,6 +332,8 @@ NSUInteger GetMaxSampleRate(const webrtc::H264ProfileLevelId &profile_level_id) 
 
   webrtc::H264BitstreamParser _h264BitstreamParser;
   std::vector<uint8_t> _frameScaleBuffer;
+
+  CMTime _previousPresentationTimeStamp;
 }
 
 // .5 is set as a mininum to prevent overcompensating for large temporary
@@ -347,6 +350,7 @@ NSUInteger GetMaxSampleRate(const webrtc::H264ProfileLevelId &profile_level_id) 
     _packetizationMode = RTCH264PacketizationModeNonInterleaved;
     _profile_level_id =
         webrtc::ParseSdpForH264ProfileLevelId([codecInfo nativeSdpVideoFormat].parameters);
+    _previousPresentationTimeStamp = kCMTimeZero;
     RTC_DCHECK(_profile_level_id);
     RTC_LOG(LS_INFO) << "Using profile " << CFStringToString(ExtractProfile(*_profile_level_id, _mode == RTCVideoCodecModeScreensharing));
     RTC_CHECK([codecInfo.name isEqualToString:kRTCVideoCodecH264Name]);
@@ -375,7 +379,9 @@ NSUInteger GetMaxSampleRate(const webrtc::H264ProfileLevelId &profile_level_id) 
   // We can only set average bitrate on the HW encoder.
   _targetBitrateBps = settings.startBitrate * 1000;  // startBitrate is in kbps.
   _bitrateAdjuster->SetTargetBitrateBps(_targetBitrateBps);
-  _encoderFrameRate = MIN(settings.maxFramerate, _maxAllowedFrameRate);
+  _targetFrameRate = MIN(settings.maxFramerate, _maxAllowedFrameRate);
+  _encoderBitrateBps = 0;
+  _encoderFrameRate = 0;
   if (settings.maxFramerate > _maxAllowedFrameRate && _maxAllowedFrameRate > 0) {
     RTC_LOG(LS_WARNING) << "Initial encoder frame rate setting " << settings.maxFramerate
                         << " is larger than the "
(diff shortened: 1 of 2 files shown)


sdk/objc/components/video_codec/RTCVideoEncoderH264.mm
```diff
@@ -365,17 +403,29 @@ - (NSInteger)startEncodeWithSettings:(RTC_OBJC_TYPE(RTCVideoEncoderSettings) *)s
...
   _maxAllowedFrameRate = static_cast<uint32_t>(GetMaxSampleRate(*_profile_level_id) /
                                                (aligned_width * aligned_height));
 
   // We can only set average bitrate on the HW encoder.
-  _targetBitrateBps = settings.startBitrate * 1000;  // startBitrate is in kbps.
-  _bitrateAdjuster->SetTargetBitrateBps(_targetBitrateBps);
-  _encoderFrameRate = MIN(settings.maxFramerate, _maxAllowedFrameRate);
+  if (_encodeMode == Constant) {
+    _targetBitrateBps = _maxBitrate;
+    // _bitrateAdjuster->SetTargetBitrateBps(_targetBitrateBps);
```
davidzhao: nit: this can be removed?

sdk/objc/components/video_codec/RTCVideoEncoderH264.mm
```diff
@@ -396,8 +446,15 @@ - (NSInteger)encode:(RTC_OBJC_TYPE(RTCVideoFrame) *)frame
   if (!_callback || !_compressionSession) {
     return WEBRTC_VIDEO_CODEC_UNINITIALIZED;
   }
-  BOOL isKeyframeRequired = NO;
 
+  CMTime presentationTimeStamp = CMTimeMake(frame.timeStampNs / rtc::kNumNanosecsPerMillisec, 1000);
+  if (CMTimeCompare(presentationTimeStamp, _previousPresentationTimeStamp) == 0) {
```
davidzhao, line 448: when does this happen? Is it an optimization for screenshare?

sdk/objc/components/video_codec/RTCVideoEncoderH264.mm
```diff
@@ -811,7 +910,12 @@ - (void)frameWasEncoded:(OSStatus)status
     RTC_LOG(LS_ERROR) << "Encode callback failed";
     return;
   }
-  _bitrateAdjuster->Update(frame.buffer.length);
+
+  // if (_encodeMode == Constant) {
```
davidzhao: is this because we've deprecated bitrateAdjuster? good to remove then?

sdk/objc/components/video_codec/RTCVideoEncoderH264.mm
```diff
@@ -54,14 +54,42 @@ - (void)frameWasEncoded : (OSStatus)status flags : (VTEncodeInfoFlags)infoFlags
 // The ratio between kVTCompressionPropertyKey_DataRateLimits and
 // kVTCompressionPropertyKey_AverageBitRate. The data rate limit is set higher
 // than the average bit rate to avoid undershooting the target.
-const float kLimitToAverageBitRateFactor = 1.5f;
+const float kLimitToAverageBitRateFactor = 10.0f;
```
davidzhao, line 57: any idea why this value had to be set so high compared to what it was before? it seems that 10x average bitrate would seem quite high, no?

davidzhao (approved):
looks good! (for the parts I was able to understand). I'm still not quite sure what the bug was that prevented it from maintaining bitrate previously.

sdk/objc/components/video_codec/RTCVideoEncoderH264.mm
cloudwebrtc: Just did some tests for camera capture, and the bit rate will be close to the target value (but not exceeding the limit). the moving picture is smooth and clear, and the bit rate of the static picture can be reduced to 1/5 of the target bit rate

cloudwebrtc: The game screen may be due to complex textures, at least 4~6mbps needs to be set to achieve smooth and clear 720p/30fps, and the camera only needs 2mbps, so this PR is good for me

sdk/objc/components/video_codec/RTCVideoEncoderH264.mm
cloudwebrtc: I tested two days ago only modifying kLimitToAverageBitRateFactor = 1.5f; to
kLimitToAverageBitRateFactor = 5.0f; also seems to get good results (keep bitrateAdjuster). will do a comparison test soon

cloudwebrtc: @hiroshihorie , can you try just modifying `kLimitToAverageBitRateFactor = 5.0f` based on the original code?

sdk/objc/components/video_codec/RTCVideoEncoderH264.mm
hiroshihorie: This is extra safety to avoid this issue:
https://developer.apple.com/forums/thread/702891

sdk/objc/components/video_codec/RTCVideoEncoderH264.mm
hiroshihorie: @davidzhao  Yes I think it's ok to remove _bitrateAdjuster.
@cloudwebrtc  kLimitToAverageBitRateFactor 5~8 seems fine but, I think chrome was using 10.0f if my memory was correct. Should we just follow what chrome uses ?

sdk/objc/components/video_codec/RTCVideoEncoderH264.mm
hiroshihorie: @davidzhao This is for computing data rate limits and not the average bitrate.

sdk/objc/components/video_codec/RTCVideoEncoderH264.mm
davidzhao: thanks! looks like this is part of reason why it's unable to keep up.

sdk/objc/components/video_codec/RTCVideoEncoderH264.mm
davidzhao: got it, I don't really understand this so feel free to ignore :)

cloudwebrtc (approved):
lgtm

sdk/objc/components/video_codec/RTCVideoEncoderH264.mm
cloudwebrtc: > @cloudwebrtc kLimitToAverageBitRateFactor 5~8 seems fine but, I think chrome was using 10.0f if my memory was correct. Should we just follow what chrome uses ?

following the chrome should be fine, and test results are as expected, so great work @hiroshihorie

hiroshihorie:
Chrome has both CBR mode and VBR(ABR) mode, not sure in what condition that switches. Will take a quick look.

sdk/objc/native/src/objc_video_track_source.mm
```diff
@@ -36,8 +36,14 @@ - (void)capturer:(RTC_OBJC_TYPE(RTCVideoCapturer) *)capturer
...
 
+#if defined(WEBRTC_ANDROID)
+constexpr int kResolutionAlignment = 16;
+#else
+constexpr int kResolutionAlignment = 2;
+#endif
+
 ObjCVideoTrackSource::ObjCVideoTrackSource(bool is_screencast)
-    : AdaptedVideoTrackSource(/* required resolution alignment */ is_screencast? 16 : 2),
+    : AdaptedVideoTrackSource(/* required resolution alignment */ kResolutionAlignment),
```
hiroshihorie: @cloudwebrtc If this looks good, I will merge this PR.

hiroshihorie: This was a misunderstanding.

PR #75, opened 2023-06-12
Fix video encoder not resuming correctly upon foregrounding

cloudwebrtc:


hiroshihorie:
Interesting! Is this required only for m114 ? does it currently exist ?

cloudwebrtc:
yeah, This is the issue mentioned by @davidliu (no issue in m104) . After iOS enters the background, it will cause the pixelBufferPool to return nil. then returning to the foreground, in m114 it does not try to restart the VT H264Encoder but directly calls RequestEncoderSwitch to switch to VP8, causing the SFU to receive the wrong payload.

m104:
https://github.com/webrtc-sdk/webrtc/blob/m104_release/video/video_stream_encoder.cc#L1891-L1905

m114:
https://github.com/webrtc-sdk/webrtc/blob/m114_release/video/video_stream_encoder.cc#L2041-L2049

hiroshihorie (approved):
Got it! Great work !

cloudwebrtc:
cherry-pick in https://github.com/webrtc-sdk/webrtc/pull/76

PR #90, opened 2023-09-08
Add scalabilityMode support for AV1/VP9.

cloudwebrtc:
- [x] iOS/macOS
- [x] Android

### DIFF
--- a/api/video_codecs/video_encoder_factory.h
+++ b/api/video_codecs/video_encoder_factory.h
@@ -85,13 +85,8 @@ class VideoEncoderFactory {
   virtual CodecSupport QueryCodecSupport(
       const SdpVideoFormat& format,
       absl::optional<std::string> scalability_mode) const {
-    // Default implementation, query for supported formats and check if the
-    // specified format is supported. Returns false if scalability_mode is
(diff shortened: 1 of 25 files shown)


api/video_codecs/video_encoder_factory.h
```diff
@@ -89,7 +89,7 @@ class VideoEncoderFactory {
     // specified format is supported. Returns false if scalability_mode is
     // specified.
     CodecSupport codec_support;
-    if (!scalability_mode) {
```
davidzhao, line 92: wouldn't this return not supported when scalability mode is unset? should the if check be removed instead?

davidzhao (commented):
lg! there's just one change that I'm not clear about

api/video_codecs/video_encoder_factory.h
cloudwebrtc: you are right, is_supported must always be returned here otherwise it will affect other encoders where scalability_mode is not set

davidzhao (approved):
lgtm!

lukasIO (approved):

PR #92, opened 2023-09-12
Use `UIInterfaceOrientation` instead of `UIDeviceOrientation` for `RTCVideoFrame`'s rotation

hiroshihorie:
`UIDeviceOrientation` contains states such as .faceUp, .faceDown which are not ideal for determining capture rotation and causes undefined behavior. This patch will use interface orientation instead.
Especially when the interface rotation is locked (by user) it had an undesirable behavior.

Also fixes issue : https://github.com/livekit/client-sdk-swift/issues/146


davidzhao (approved):
lgtm

PR #93, opened 2023-09-14
fix h264 freeze.

cloudwebrtc:


lukasIO (approved):
lgtm!

theomonnom (approved):
lgtm! TIL about emulation_prevention_three_byte

api/crypto/frame_crypto_transformer.cc
```diff
@@ -82,6 +82,28 @@ const EVP_CIPHER* GetAesCbcAlgorithmFromKeySize(size_t key_size_bytes) {
...
+      return videoFrame->header().codec ==
+             webrtc::VideoCodecType::kVideoCodecH264;
+    }
+    default:
+      return false;
+  }
+}
+
+inline bool NeedsRbspUnescaping(const uint8_t* frameData, size_t frameSize) {
+  for (size_t i = 0; i < frameSize - 3; ++i) {
```
theomonnom, line 100:
```suggestion
  for (size_t i = 0; i < frameSize - 2; ++i) {
```

theomonnom (changes requested):

api/crypto/frame_crypto_transformer.cc
cloudwebrtc: I want to say that you should be right, but it seems that rbsp will not appear at the end of the frame. Because 003 is to escape 0001, such as 0001 change to 00301,

PR #95, opened 2023-09-20
Fix/send frame cryptor events from signaling thread

cloudwebrtc:
* replace all `std::shared_ptr` to `rtc::scoped_refptr`
* send encryption state events from the `signaling-thread`
* create framecryptor with signaling_thread. (break changes)

### DIFF
--- a/api/crypto/frame_crypto_transformer.cc
+++ b/api/crypto/frame_crypto_transformer.cc
@@ -342,7 +342,7 @@ void FrameCryptorTransformer::encryptFrame(
         << "FrameCryptorTransformer::encryptFrame() sink_callback is NULL";
     if (last_enc_error_ != FrameCryptionState::kInternalError) {
       last_enc_error_ = FrameCryptionState::kInternalError;
-      onFrameCryptionStateChanged(last_dec_error_);
+      onFrameCryptionStateChanged(last_enc_error_);
     }
     return;
   }
@@ -364,7 +364,7 @@ void FrameCryptorTransformer::encryptFrame(
                      << participant_id_;
     if (last_enc_error_ != FrameCryptionState::kMissingKey) {
       last_enc_error_ = FrameCryptionState::kMissingKey;
-      onFrameCryptionStateChanged(last_dec_error_);
+      onFrameCryptionStateChanged(last_enc_error_);
     }
     return;
   }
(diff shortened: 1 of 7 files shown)


api/crypto/frame_crypto_transformer.cc
```diff
@@ -341,9 +342,7 @@ void FrameCryptorTransformer::encryptFrame(
         << "FrameCryptorTransformer::encryptFrame() sink_callback is NULL";
     if (last_enc_error_ != FrameCryptionState::kInternalError) {
       last_enc_error_ = FrameCryptionState::kInternalError;
-      if (observer_)
-        observer_->OnFrameCryptionStateChanged(participant_id_,
-                                               last_enc_error_);
+      onFrameCryptionStateChanged(last_dec_error_);
```
boks1971: sorry, just showed up in my threads, just looking out of curiosity, should this the next three be `last_enc_error_`?

api/crypto/frame_crypto_transformer.h
```diff
@@ -417,10 +425,12 @@ class RTC_EXPORT FrameCryptorTransformer
  private:
   void encryptFrame(std::unique_ptr<webrtc::TransformableFrameInterface> frame);
   void decryptFrame(std::unique_ptr<webrtc::TransformableFrameInterface> frame);
+  void onFrameCryptionStateChanged(FrameCryptionState error);
   rtc::Buffer makeIv(uint32_t ssrc, uint32_t timestamp);
   uint8_t getIvSize();
 
  private:
+  TaskQueueBase* const signaling_thread_;
```
theomonnom, line 433: Is there a reason to use TaskQueueBase* const signaling_thread_; instead of a pointer to rtc::Thread?

api/crypto/frame_crypto_transformer.cc
cloudwebrtc: Ah, okay you found a critical bug! thanks. Error caused by copy/paste xD

cloudwebrtc: fixed

theomonnom (approved):
lgtm

api/crypto/frame_crypto_transformer.h
cloudwebrtc: We only use the task queue to send events, so there is no need for the entire thread object
like this one https://github.com/webrtc-sdk/webrtc/blob/m114_release/pc/peer_connection_message_handler.h#L47

theomonnom: oh ok, lg!

PR #96, opened 2023-09-21
more improvements for E2EE.

cloudwebrtc:
* Use independent threads to process frame encryption/decryption
* Correct handle SIF frame, catch up with js rules
* Fix a fault tolerance judgment failure

### DIFF
--- a/api/crypto/frame_crypto_transformer.cc
+++ b/api/crypto/frame_crypto_transformer.cc
@@ -680,12 +680,12 @@ void FrameCryptorTransformer::decryptFrame(
   }
 
   if (!decryption_success) {
-    if (last_dec_error_ != FrameCryptionState::kDecryptionFailed || ratchet_count >= key_provider_->options().ratchet_window_size) {
-      last_dec_error_ = FrameCryptionState::kDecryptionFailed;
       if(key_handler->DecryptionFailure()) {
-        onFrameCryptionStateChanged(last_dec_error_);
+        if (last_dec_error_ != FrameCryptionState::kDecryptionFailed) {
+          last_dec_error_ = FrameCryptionState::kDecryptionFailed;
+          onFrameCryptionStateChanged(last_dec_error_);
+        }
       }
-    }
     return;
   }
 
--- a/api/crypto/frame_crypto_transformer.cc
+++ b/api/crypto/frame_crypto_transformer.cc
@@ -314,11 +314,18 @@ FrameCryptorTransformer::FrameCryptorTransformer(
     Algorithm algorithm,
     rtc::scoped_refptr<KeyProvider> key_provider)
     : signaling_thread_(signaling_thread),
+      thread_(rtc::Thread::Create()),
       participant_id_(participant_id),
       type_(type),
       algorithm_(algorithm),
       key_provider_(key_provider) {
   RTC_DCHECK(key_provider_ != nullptr);
+  thread_->SetName("FrameCryptorTransformer", nullptr);
+  thread_->Start();
+}
+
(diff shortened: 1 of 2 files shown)


api/crypto/frame_crypto_transformer.cc
```diff
@@ -665,10 +680,11 @@ void FrameCryptorTransformer::decryptFrame(
   }
 
   if (!decryption_success) {
-    if (last_dec_error_ != FrameCryptionState::kDecryptionFailed) {
+    if (last_dec_error_ != FrameCryptionState::kDecryptionFailed || ratchet_count >= key_provider_->options().ratchet_window_size) {
```
theomonnom: Will this call multiple time the kDecryptionFailed callback?

api/crypto/frame_crypto_transformer.cc
```diff
@@ -371,8 +388,9 @@ void FrameCryptorTransformer::encryptFrame(
 
   rtc::ArrayView<const uint8_t> date_in = frame->GetData();
   if (date_in.size() == 0 || !enabled_cryption) {
-    sink_callback->OnTransformedFrame(std::move(frame));
-    return;
+      RTC_LOG(LS_WARNING)
```
theomonnom: - Will you undo this change? (When FrameCryptor is not enabled, discard the input frame)

theomonnom (changes requested):

theomonnom (approved):
otherwise. smtg

api/crypto/frame_crypto_transformer.cc
```diff
@@ -490,25 +508,22 @@ void FrameCryptorTransformer::decryptFrame(
...
+      << "FrameCryptorTransformer::decryptFrame() date_in.size() == 0 || enabled_cryption == false";
     return;
   }
 
   auto uncrypted_magic_bytes = key_provider_->options().uncrypted_magic_bytes;
   if (uncrypted_magic_bytes.size() > 0 &&
-      date_in.size() >= uncrypted_magic_bytes.size() + 1) {
+      date_in.size() >= uncrypted_magic_bytes.size()) {
     auto tmp =
-        date_in.subview(date_in.size() - (uncrypted_magic_bytes.size() + 1),
```
theomonnom, line 501: is it the only change made to sif?

api/crypto/frame_crypto_transformer.cc
cloudwebrtc: yes, I should revert this change

api/crypto/frame_crypto_transformer.cc
cloudwebrtc: There seems to be no need to judge ratchet_count here, because when decryption_success == false, either ratchet_window_size = 0, or ratchet_count exceeds the number of times，
change to this should be ok
```cpp
  if (!decryption_success) {
      if(key_handler->DecryptionFailure()) {
        if (last_dec_error_ != FrameCryptionState::kDecryptionFailed) {
          last_dec_error_ = FrameCryptionState::kDecryptionFailed;
          onFrameCryptionStateChanged(last_dec_error_);
        }
      }
    return;
  }
```

api/crypto/frame_crypto_transformer.cc
cloudwebrtc: Yes, when we discussed with @lukasIO  earlier, we reserved a 1 byte for frameType. It seems that it has been abandoned in the latest js-sdk.

lukasIO: yeah sorry, we decided on that when discussing with Raja, should have made this more explicit!

PR #100, opened 2023-10-04
Fix missing `RTC_OBJC_TYPE` macros

hiroshihorie:
Some symbols are missing `RTC_OBJC_TYPE` macro, this is required for prefixing.
Prefix will be applied by defining `RTC_OBJC_TYPE_PREFIX` in `RTCMacros.h` and compiling.

### DIFF
--- a/sdk/objc/base/RTCMacros.h
+++ b/sdk/objc/base/RTCMacros.h
@@ -38,7 +38,7 @@
 //
 // This macro must only be defined here and not on via compiler flag to
 // ensure it has a unique value.
-#define RTC_OBJC_TYPE_PREFIX
+#define RTC_OBJC_TYPE_PREFIX LK_
 
 // RCT_OBJC_TYPE
 //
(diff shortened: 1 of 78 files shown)


sdk/objc/base/RTCMacros.h
```diff
@@ -38,7 +38,7 @@
 //
 // This macro must only be defined here and not on via compiler flag to
 // ensure it has a unique value.
-#define RTC_OBJC_TYPE_PREFIX
+#define RTC_OBJC_TYPE_PREFIX LK_
```
davidzhao: this is unintended?

davidzhao (approved):
lgtm!

sdk/objc/base/RTCMacros.h
hiroshihorie: Yup! I reverted that !

cloudwebrtc (approved):

PR #105, opened 2023-12-06
Fix localhost TURN server conection issue

PonyPC:
[flutter-webrtc/flutter-webrtc/issues/#1464](https://github.com/flutter-webrtc/flutter-webrtc/issues/1464)


PonyPC:
Chrome and Edge is not the same behavior like this repository, they allow the webbrowser connect to the local turn server.
So I think the ignore type should be unknown instead of loopback

davidzhao:
We do not want to diverge here.

PR #110, opened 2024-04-02
Allow ice gathering on any address ports

davidliu:
Some VPNs use 0.0.0.0 mask which will be ignored by the default gathering.
This flag allows those to be picked up as a last resort.

Has implementations for both android and ios.


cloudwebrtc (approved):
lgtm

PR #112, opened 2024-04-25
add PrivacyInfo.xcprivacy to darwin frameworks.

cloudwebrtc:


sdk/objc/PrivacyInfo.xcprivacy
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>NSPrivacyTrackingDomains</key>
+	<array/>
+	<key>NSPrivacyAccessedAPITypes</key>
+	<array/>
```
hiroshihorie: It appears you haven't specified what type of APIs is used yet.

I think WebRTC uses [System boot time APIs](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api#4278394)
```xml
	<key>NSPrivacyAccessedAPITypes</key>
	<array>
		<dict>
			<key>NSPrivacyAccessedAPIType</key>
			<string>NSPrivacyAccessedAPICategorySystemBootTime</string>
			<key>NSPrivacyAccessedAPITypeReasons</key>
			<array>
				<string>35F9.1</string>
				<string>8FFB.1</string>
			</array>
		</dict>
	</array>
```

Does it also use [Disk space APIs](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api#4278394) ?

hiroshihorie (changes requested):

sdk/objc/PrivacyInfo.xcprivacy
hiroshihorie: It appears to also be using stat() API, used to check file/folder existence:
Screenshot 2024-04-25 10 09 27

Discussion : https://forums.developer.apple.com/forums/thread/734750

So the final result can look like:

```xml
	<key>NSPrivacyAccessedAPITypes</key>
	<array>
		<dict>
			<key>NSPrivacyAccessedAPIType</key>
			<string>NSPrivacyAccessedAPICategoryFileTimestamp</string>
			<key>NSPrivacyAccessedAPITypeReasons</key>
			<array>
				<string>C617.1</string>
			</array>
		</dict>
		<dict>
			<key>NSPrivacyAccessedAPIType</key>
			<string>NSPrivacyAccessedAPICategorySystemBootTime</string>
			<key>NSPrivacyAccessedAPITypeReasons</key>
			<array>
				<string>35F9.1</string>
				<string>8FFB.1</string>
			</array>
		</dict>
	</array>
```

hiroshihorie: I'm not confident the NSPrivacyAccessedAPITypeReasons are correct, can you double check ? Thanks !

cloudwebrtc: I compared here and the list you wrote looks right
https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api

davidliu (approved):

PR #113, opened 2024-04-25
Fix naming for yuv helper

cloudwebrtc:


sdk/objc/helpers/RTCYUVHelper.mm
```diff
@@ -27,7 +27,7 @@ + (void)I420Rotate:(const uint8_t*)srcY
               dstV:(uint8_t*)dstV
         dstStrideV:(int)dstStrideV
              width:(int)width
-             width:(int)height
+             height:(int)height
```
davidzhao, line 30: oh wow... nice catch!

davidzhao (approved):

PR #115, opened 2024-05-06
Fix initial camera rotation

hiroshihorie:
It's too late to resolve orientation after starting camera capture, the first couple of frames will have incorrect rotation information.
This patch starts listening for orientation as soon as RTCCameraVideoCapturer is created.

### DIFF
--- a/sdk/objc/components/capturer/RTCCameraVideoCapturer.m
+++ b/sdk/objc/components/capturer/RTCCameraVideoCapturer.m
@@ -43,7 +43,6 @@ const int64_t kNanosecondsPerSecond = 1000000000;
   RTCVideoRotation _rotation;
 #if TARGET_OS_IPHONE
   UIInterfaceOrientation _orientation;
-  BOOL _generatingOrientationNotifications;
 #endif
 }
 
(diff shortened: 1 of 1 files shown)


sdk/objc/components/capturer/RTCCameraVideoCapturer.m
```diff
@@ -160,17 +173,6 @@ - (void)startCaptureWithDevice:(AVCaptureDevice *)device
                     block:^{
                       RTCLogInfo("startCaptureWithDevice %@ @ %ld fps", format, (long)fps);
 
-#if TARGET_OS_IPHONE
-                      dispatch_async(dispatch_get_main_queue(), ^{
```
davidliu, line 164: Would it be enough to just change these to `dispatch_sync` to ensure the `updateOrientation` gets called? Seems still possible for the threading to get ordered such that the `updateOrientation` call comes in late, even if unlikely.

Also would mean that the capturer doesn't have to turn on the accelerometer all the time.

hiroshihorie: Yes calling `beginGeneratingDeviceOrientationNotifications()` when not necessary isn't ideal, will make changes.

hiroshihorie:
Will close this for now.

PR #117, opened 2024-05-17
Strip unused field trial global var

hiroshihorie:


davidzhao (approved):
much better!

davidliu (approved):

hiroshihorie:
After more digging, I think I can't simply strip this out.
static var is used to keep string in memory.

https://github.com/webrtc-sdk/webrtc/blob/7cf1e433027c52bd8f0e66254f716b56eeeb9fb5/system_wrappers/include/field_trial.h#L85

https://github.com/webrtc-sdk/webrtc/blob/7cf1e433027c52bd8f0e66254f716b56eeeb9fb5/examples/peerconnection/client/main.cc#L86-L87

PR #118, opened 2024-05-19
android: make audio output attributes modifiable

davidliu:
Needed to do audio mode changes on react native


cloudwebrtc (approved):
lgtm!

PR #121, opened 2024-05-28
Set RTCCameraVideoCapturer initial zoom factor

hiroshihorie:
Some devices, such as `.builtInTripleCamera`, have a zoom factor of 2.0 for the normal zoom. This adjustment changes it to 2.0 instead of 1.0. Without this adjustment, when using `.builtInTripleCamera`, it will appear zoomed out.


cloudwebrtc (approved):
lgtm!

PR #127, opened 2024-06-24
Fix AudioDevice delegate

npazkevich:


npazkevich:
Sorry, it's been made by mistake

PR #132, opened 2024-07-15
Allow to pass in capture session to RTCCameraVideoCapturer

hiroshihorie:
Expose initializers to pass in capture session to RTCCameraVideoCapturer so we can use AVCaptureMultiCamSession etc to capture front and back simultaneously for iOS.


cloudwebrtc (approved):
lgtm

rajleechaudry:
Is there a way to use this readily from react-native?

PR #134, opened 2024-07-15
fix: correctly map selected layers to simulcast layers

### LINKED ISSUE #133: Low resolution on some Android phones
Hi there,

during development of our app that uses Livekit as the video call framework, we discovered that the video quality is very low on some Android devices. I tracked down at least one bug (and a combination of heuristics combined with some bad luck I guess) in WebRTC, so I'm reporting this here and not in https://github.com/livekit/client-sdk-flutter/, the entry point we use.

## The issue
We set the `defaultCameraCaptureOptions` to `VideoParametersPresets.h720_43` (i.e. `960x720`) with simulcast enabled. On some Android phones, the sent simulcast layers are `144x192` and `288x384`, even on local network with a local livekit server. This is ~16% of the requested resolution, and definitely not suitable for high quality video conferencing.

## The deep dive into WebRTC
After ruling out network issues, I started a deep dive into client-sdk-flutter and subsequently into WebRTC itself. Here are my findings (on a Fairphone 5 with Android 13):
- Livekit room options:
```dart
RoomOptions(
  adaptiveStream: true,
  dynacast: true,
  defaultAudioCaptureOptions:
      AudioCaptureOptions(deviceId: _settings.audioDeviceId),
  defaultCameraCaptureOptions: CameraCaptureOptions(
      deviceId: _settings.videoDeviceId,
      params: VideoParametersPresets.h720_43
)));
```
- Camera2Session lists following available capture formats: `[4080x3060, 4000x3000, 4000x2250, 3960x1760, 3840x2160, 3264x2448, 3072x3072, 3060x3060, 2700x1200, 2592x1944, 2048x1536, 1920x1080, 1600x1200, 1440x1080, 1280x960, 1280x720, 1200x1200, 1080x1080, 1024x768, 800x600, 800x480, 720x480, 640x480, 640x360, 352x288, 320x240, 176x144]`
- based on the requested resolution of 960x720, Camera2Session uses the nearest (in terms of total pixels) available format, namely 1024x768: `Camera2Session: Using capture format: 1024x768@[7.0:30.0]`
- The first major reduction of resolution happens on the boundary between Java and C++: https://github.com/webrtc-sdk/webrtc/blob/7ddfc43fcd2e4ba0548258a76fe88d49b34588c5/media/base/video_adapter.cc#L234-L236
  - Using the requested pixel count (960x720=691200) as maximum, the video adapter tries to get the input video (1024x768=786432) below that maximum. It only uses `2/3` and `3/4` as factors for efficiency reasons. 3/4 on each axis reduces the number of overall pixels to 9/16 which is ~56%. The current resolution is now 768x576
- This downscaling causes simulcast to only send 2 layers instead of 3 because it doesn't reach the 960*540 bucket anymore: https://github.com/webrtc-sdk/webrtc/blob/7ddfc43fcd2e4ba0548258a76fe88d49b34588c5/video/config/simulcast.cc#L82-L106
- The simulcast layers are sorted by ascending resolution and originally have scale factors 4, 2 and 1
- Finally, due to a bug in the EncoderStreamFactory, the two remaining layers (768x576 and 384*288) are further reduced to 384x288 and 192x144, because they are mapped to the scale factors 4 and 2, not 2 and 1: https://github.com/webrtc-sdk/webrtc/blob/7ddfc43fcd2e4ba0548258a76fe88d49b34588c5/video/config/encoder_stream_factory.cc#L358

## The solution(s)
- one of
  - don't allow Camera2Session to take larger capture format than requested to prevent VideoAdapter to half the video stream later on
  - allow the VideoAdapter a slightly increased maximum resolution, similar to `max_roundup_rate` in simulcast layer selection
- maybe add 4:3 tables for simulcast layer selection to compensate for the lower resolution of 4:3 videos compared to 16:9 videos for same height. Currently 4:3 videos tend to fall into lower buckets (with lower layer count)
- fix layer mapping bug (PR follows, EDIT here it is: #134)

## one more thing
With and without the bug fix, livekit server reports a "Good" connection quality for that phone despite being on the same network. On a different phone, that has 960x720 in its native capture formats, layer selection works and connection quality is "Excellent". This is likely a bug in livekits scorer that I will investigate separately. I still wanted to mention it here because it was the reason for my investigations

holzgeist:
this partly fixes #133.
Quick summary:
- on lower resolutions, simulcast only sends 2 (or even 1) layer.
- The layers are ordered from low to high resolution (downscale factors 4, 2, 1)
- previously, when only sending 2 layers (factors 2, 1), the layer with factor 2 was mapped to the layer with factor 4, i.e. scaled down again)
- with this patch, the layers are mapped to the correct factor and not scaled down again

### DIFF
--- a/video/config/encoder_stream_factory.cc
+++ b/video/config/encoder_stream_factory.cc
@@ -354,71 +354,74 @@ EncoderStreamFactory::CreateSimulcastOrConferenceModeScreenshareStreams(
           ? NormalizeSimulcastSize(trials_, height,
                                    encoder_config.number_of_streams)
           : height;
+
   for (size_t i = 0; i < layers.size(); ++i) {
-    layers[i].active = encoder_config.simulcast_layers[i].active;
-    layers[i].scalability_mode =
-        encoder_config.simulcast_layers[i].scalability_mode;
-    layers[i].requested_resolution =
-        encoder_config.simulcast_layers[i].requested_resolution;
+    // account for layers removed in `GetSimulcastConfig`. Layers are ordered
+    // from lowest to highest quality, including their
+    // `scale_down_resolution_by` factor. So we need to map the last
+    // simulcast_layer to the last layer, not the other way round. Otherwise the
+    // scaling factor is applied multiple times
+
+    size_t simulcast_layer_index =
+        i + std::max(1, encoder_config.simulcast_layers.size() - layers.size());
+    auto& simulcast_layer = encoder_config.simulcast_layers[simulcast_layer_i];
+    layers[i].active = simulcast_layer.active;
+    layers[i].scalability_mode = simulcast_layer.scalability_mode;
+    layers[i].requested_resolution = simulcast_layer.requested_resolution;
     // Update with configured num temporal layers if supported by codec.
-    if (encoder_config.simulcast_layers[i].num_temporal_layers &&
+    if (simulcast_layer.num_temporal_layers &&
         IsTemporalLayersSupported(codec_name_)) {
-      layers[i].num_temporal_layers =
-          *encoder_config.simulcast_layers[i].num_temporal_layers;
+      layers[i].num_temporal_layers = *simulcast_layer.num_temporal_layers;
     }
-    if (encoder_config.simulcast_layers[i].max_framerate > 0) {
-      layers[i].max_framerate =
-          encoder_config.simulcast_layers[i].max_framerate;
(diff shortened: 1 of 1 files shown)


davidliu (changes requested):
Don't think we can use this PR, as noted in the related issue:

https://github.com/webrtc-sdk/webrtc/issues/133#issuecomment-2239082153

PR #135, opened 2024-07-17
Fix NetworkMonitor race condition when dispatching native observers

davidliu:
There is a race condition in NetworkMonitor where native observers may be removed concurrently with a notification being dispatched, leading to a dangling pointer dereference (trying to dispatch an observer that was already removed and destroyed), and from there a crash with access violation.

By ensuring dispatching to native observers is done within the synchronization lock that guards additions/removals of native observers protects against this race condition. Since native observers callbacks are posted to the networking thread in the C++ side anyway, there should be no risk of deadlock/starvation due to long-running observers.

Bug: webrtc:15837
Change-Id: Id2b788f102dbd25de76ceed434c4cd68aa9a569e
Reviewed-on: https://webrtc-review.googlesource.com/c/src/+/338643
Reviewed-by: Taylor Brandstetter <deadbeef@webrtc.org>
Commit-Queue: Harald Alvestrand <hta@webrtc.org>
Reviewed-by: Harald Alvestrand <hta@webrtc.org>
Cr-Commit-Position: refs/heads/main@{#42256}

### DIFF
--- a/sdk/android/api/org/webrtc/NetworkMonitor.java
+++ b/sdk/android/api/org/webrtc/NetworkMonitor.java
@@ -227,10 +227,13 @@ public class NetworkMonitor {
   /** Alerts all observers of a connection change. */
   private void notifyObserversOfConnectionTypeChange(
       NetworkChangeDetector.ConnectionType newConnectionType) {
-    List<Long> nativeObservers = getNativeNetworkObserversSync();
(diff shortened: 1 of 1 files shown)


davidliu:
Original bug: https://issues.webrtc.org/issues/42226201

cloudwebrtc (approved):

PR #138, opened 2024-08-14
Remove bssl::ScopedXXX for frame cryptor.

cloudwebrtc:
to support both boringssl and openssl.


cloudwebrtc:
Not needed at the moment, but as a backup

cloudwebrtc:
Using the `bssl::Scoped` pointer template to restrict users to only use boringssl may be better for security, because when supporting `openssl` users may introduce vulnerabilities by switching to an insecure version of `openssl`.

PR #140, opened 2024-09-01
Add isDisposed to MediaStreamTrack

davidliu:


sdk/android/api/org/webrtc/MediaStreamTrack.java
```diff
@@ -121,6 +121,10 @@ private void checkMediaStreamTrackExists() {
     }
   }
 
+  public boolean isDisposed() {
```
maitrungduc1410, line 124: what's the usage of this flag?

PR #156, opened 2024-11-26
Use `rtc::ToString` instead of `std::to_string` in `SocketAddress::PortAsString()`

mgsloan:
Justification for this change is that `std::to_string` should be avoided as it uses the user's locale and calls to it get serialized, which is bad for concurrency.

My actual motivation for this is quite bizarre. Before this change, with Zed's use of the LiveKit Rust SDK, I was getting connection strings that were not valid utf-8, instead having a port of `3\u0000\u0000\u001c\u0000`. I have not figured out how that could happen or why this change fixes it.


theomonnom (approved):

cloudwebrtc (approved):
lg!

PR #171, opened 2025-05-20
Prefix enums with `RTC_OBJC_TYPE` macro

hiroshihorie:
Continued from https://github.com/webrtc-sdk/webrtc/pull/100 also prefix enums.
I think this shouldn't break compiling.


cloudwebrtc (approved):
lgtm

PR #174, opened 2025-06-25
Fix missing `RTC_OBJC_TYPE` macro

harshs-dyte:
Specifying the `rtc_objc_prefix` flag fails the build for these symbols since the macro wasn't applied to them.


hiroshihorie (approved):
Oops missed that one! Thanks ✅

PR #176, opened 2025-06-30
fix(generate_licenses): add missing licenses for ffmpeg and openh264

harshs-dyte:
Doing a custom build with the following command line fails the build during the license generation step because these components were added to the build without their licenses being declared here.

`python tools_webrtc/ios/build_ios_libs.py --build_config release --arch arm64 x64 --extra-gn-args ios_enable_code_signing=false is_component_build=false rtc_include_tests=false is_debug=false enable_dsyms=false rtc_libvpx_build_vp9=true use_goma=false rtc_enable_symbol_export=true rtc_build_examples=false rtc_use_h264=true use_rtti=true use_custom_libcxx=false`

### DIFF
--- a/tools_webrtc/libs/generate_licenses.py
+++ b/tools_webrtc/libs/generate_licenses.py
@@ -48,6 +48,7 @@ LIB_TO_LICENSES_DICT = {
         'third_party/android_deps/libs/'
         'com_google_errorprone_error_prone_core/LICENSE'
     ],
+    'ffmpeg': ['third_party/ffmpeg/LICENSE.md'],
     'fiat': ['third_party/boringssl/src/third_party/fiat/LICENSE'],
     'guava': ['third_party/android_deps/libs/com_google_guava_guava/LICENSE'],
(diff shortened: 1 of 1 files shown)


harshs-dyte:
I see there's a patch [in the webrtc-build repo](https://github.com/webrtc-sdk/webrtc-build/blob/f8ed4eff3bda836fe9c44d5c323ae9f77f314f3f/build/patches/add_license_dav1d.patch) adding licenses, this should probably go there instead?

cloudwebrtc (approved):

harshs-dyte:
Anything else I need to do to allow this to be merged?

davidliu:
@harshs-dyte hey, we just updated to m137 and switched branches. Would it be possible to open the PR against `m137_release`?

harshs-dyte:
> @harshs-dyte hey, we just updated to m137 and switched branches. Would it be possible to open the PR against `m137_release`?

Sure thing.

PR #181, opened 2025-07-22
fix: Fixed scalabilityModes is invalid on Apple platforms.

cloudwebrtc:
fixed VP9/AV1 encoder unusable.

We have to apply this patch and publish new binary packages. Otherwise we can't use AV1/VP9 codec in Apple platform.


hiroshihorie (approved):

hiroshihorie:
Did we miss this at merge of m137 ?

cloudwebrtc:
I think this change has been merged into m137_release. and should be included in the latest build.

PR #191, opened 2025-08-25
feat: Data packet cryptor.

cloudwebrtc:
TODO:
- [x] Core Implementation
- [x] Android JNI Wrapper
- [x] Objective-C Wrapper


api/crypto/frame_crypto_transformer.cc
```diff
@@ -753,4 +753,43 @@ uint8_t FrameCryptorTransformer::getIvSize() {
...
+    const webrtc::scoped_refptr<EncryptedPacket> encryptedPacket) {}
+
+rtc::Buffer DataPacketCryptor::makeIv(uint32_t timestamp) {
+  if (send_count_ = 0) {
+    srand((unsigned)time(NULL));
+    send_count_ = floor(rand() * 0xFFFF);
+  }
+  rtc::ByteBufferWriter buf;
+  srand((unsigned)time(NULL));
+  uint32_t random_u32 = rand();
```
davidliu: I think we should use `CreateRandomId()` from [crypto_random.h](https://github.com/webrtc-sdk/webrtc/blob/m137_release/rtc_base/crypto_random.h). `rand()` is not cryptographically secure. I noticed we similarly use `rand()` above [here](https://github.com/webrtc-sdk/webrtc/pull/191/files#diff-7c5d6536fc3fee31501ca8f2f430869b0cac2a050abefe6b5077af1d92ae2068R732) which should also be changed.

api/crypto/frame_crypto_transformer.cc
```diff
@@ -753,4 +761,156 @@ uint8_t FrameCryptorTransformer::getIvSize() {
   }
 }
 
+DataPacketCryptor::DataPacketCryptor(
```
davidliu, line 764: Hey, I noticed this doesn't use the signalling thread. Will this cause any multi-threading issues with the key provider potentially being used at the same time as frame crypting?

cloudwebrtc: I'm not sure. The signaling thread in FrameCryptor is actually only used to deliver the [onFrameCryptionStateChanged](https://github.com/webrtc-sdk/webrtc/blob/duan/data-packet-cryptor/api/crypto/frame_crypto_transformer.cc#L725-L735) event. Internally, a [separate thread ["FrameCryptorTransformer"]](https://github.com/webrtc-sdk/webrtc/blob/duan/data-packet-cryptor/api/crypto/frame_crypto_transformer.cc#L353) is used to [encryptFrame/decryptFrame](https://github.com/webrtc-sdk/webrtc/blob/duan/data-packet-cryptor/api/crypto/frame_crypto_transformer.cc#L374-L382). Unless we use same thread as a worker thread for DataPacketCryptor to avoid keyProvider contention.

cloudwebrtc: I don't think they compete strongly by default. It only happens once when retrieving the key_handler `key_handler->GetKeySet(key_index);` and once when the key ratchet needs to update the key material. 
This is essentially similar to the two thread competition in frameCryptor for rtp_sender and rtp_receiver.

PR #178, opened 2025-07-10
m137 AVAudioEngine ver ADM

hiroshihorie:

PR #205, opened 2025-11-17
Revert default AudioOptions change for webrtc_voice_engine.cc on Android

davidliu:
These default values got changed from true to false in #178. On Android, the way the audio options are set is fairly janky, and actually relied on these initially defaulting to true.

I'm not sure if other non iOS platforms rely on these initial default values as well, but that's something to be aware of for those platforms.


davidliu:
Build: https://github.com/webrtc-sdk/webrtc-build/actions/runs/19435095034

hiroshihorie (approved):

PR #209, opened 2025-12-20
鼠标显示与隐藏

110jiangnan:
macOS下 新增录屏鼠标的显示与隐藏，这些改动我没有测试，但是编译是可以通过的，需要官方来进行测试。


hiroshihorie:
Thank you for the PR, I did a little investigation and it looks like we can set the cursor visibility through options at run-time. I think that's a better route . I will double check.

PR #215, opened 2026-01-27
Try to fix Mac Catalyst crash on disconnect

pblazej:
Detach `AVAudioEngine` nodes before calling stop to invalidate audio callbacks and prevent race condition where callbacks access the aggregate device after it's destroyed.

Tentative fix for: https://github.com/livekit/client-sdk-swift/issues/864?

### DIFF
--- a/modules/audio_device/audio_engine_device.mm
+++ b/modules/audio_device/audio_engine_device.mm
@@ -1801,6 +1801,55 @@ int32_t AudioEngineDevice::ApplyDeviceEngineState(EngineStateUpdate state) {
       configuration_observer_ = nil;
     }
 
+    // Detach all nodes before stopping engine to invalidate audio callbacks.
+    // This prevents race conditions where callbacks access the aggregate device
+    // after it's been destroyed (fixes Mac Catalyst crash on disconnect).
+    LOGI() << "Detaching nodes before engine stop...";
+
+    // Detach source node (output)
+    if (source_node_ != nil) {
+      @try {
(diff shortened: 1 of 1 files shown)


hiroshihorie:
Nice, were you able to reproduce the crash / fix ?

modules/audio_device/audio_engine_device.mm
```diff
@@ -1801,6 +1801,55 @@ AVAudioVoiceProcessingOtherAudioDuckingLevel ToAVDuckingLevel(
...
+      input_mixer_node_ = nil;
+    }
+
+    // Dispose audio converter used by sink callback (Float32 -> Int16).
+    // Must be done after detaching sink node to avoid race with callback.
+    if (converter_ref_ != nullptr) {
+      AudioConverterDispose(converter_ref_);
+      converter_ref_ = nullptr;
+    }
+    converter_buffer_ = nil;
```
hiroshihorie, line 1851: Is this logic redundant, since it's disposed in the flow below ?

hiroshihorie:
I have a feeling this creates other issues, since if we detach nodes on a stop, when we start again the nodes are detached.
Stop -> Start (Restart) can happen for interruptions or MuteMode:Restart.

But if this fixes the macOS catalyst crash, that's interesting ...

Alternatively we can shift/move the whole stop block to a later timing below, after the "disable input/output" logic.

pblazej:
@hiroshihorie I think it needs exact reproduction steps still, can do more harm than good.

PR #173, opened 2025-06-12
Update to m137

pblazej:
I changed my strategy and merged old → new, not as clear, but the amount of conflicts in the Obj-C++ part (especially around prefixing) made it unmanageable.

### Apple
- 🟢 iOS - works
- 🟢 macOS - works
- 🟢 Catalyst - works
- 🟡 tvOS - runs
- 🟡 xrOS - runs, requires 2.0 ⚠️ 
```
ld64.lld: error: ../../third_party/llvm-build/Release+Asserts/lib/clang/21/lib/darwin/libclang_rt.xros.a(addtf3.c.o) has version 2.2.0, which is newer than target minimum of 1.1.0
```

### Android
- 🟢 works

### Flutter
- 🟢 works

### Tests
- unknown

## Other PRs
- https://github.com/webrtc-sdk/build/pull/3
- https://github.com/webrtc-sdk/webrtc-build/pull/32

PR #217, opened 2026-01-29
feat(m144 upgrade): patch from m137_release (#173)

cloudwebrtc:


pblazej:
🍏 

- One little crash (I guess it happened before): [`7e276d8` (this PR)](https://github.com/webrtc-sdk/webrtc/pull/217/commits/7e276d82cc967838a8b35fc368c109377f715f66)
- Some crypto errors after key ratcheting - looks like it's expected (again) for the shared key https://github.com/livekit/client-sdk-swift/pull/781#issuecomment-3284076499

sdk/objc/api/peerconnection/RTCPeerConnectionFactory.mm
```diff
@@ -54,19 +65,31 @@
...
+#import "components/audio/RTCAudioProcessingModule.h"
+#import "components/audio/RTCDefaultAudioProcessingModule+Private.h"
+
 #if defined(WEBRTC_IOS)
 #import "sdk/objc/native/api/audio_device_module.h"
 #endif
 
+static webrtc::Environment CreateDefaultEnvironment() {
+  return webrtc::CreateEnvironment(
+      std::make_unique<webrtc::FieldTrials>("WebRTC-IceHandshakeDtls/Enabled/"));
```
pblazej: I think there's no way to pass it on Apple vs Kotlin `RTCModule.kt via PeerConnectionFactory.Builder.setFieldTrials()`, maybe we should invent one to avoid hardcoding?

cc @davidliu

pblazej:
🍏 

AV1 encoder is failing:
```
(libaom_av1_encoder.cc:985): LibaomAv1Encoder::Encode returned error: 'Invalid parameter'.
```

The hypothesis is dimensions mismatch, but I could not reproduce that with VP (same cases):
```
AV1 is the only codec that wraps its image at the buffer's dimensions instead of the config's dimensions.
```

Logs:
```
(libaom_av1_encoder.cc:1040): aom_codec_encode failed: Invalid parameterImage size must match encoder init configuration size img=1280x720 cfg=640x360 pts=2225161 dur=3000
(libaom_av1_encoder.cc:985): LibaomAv1Encoder::Encode returned error: 'Invalid parameter'.
(video_stream_encoder.cc:2225): Failed to encode frame. Error code: -1
```

Upstream commit:
9eca82bdb2 ("Pass lower resolution image to libaom av1 encoder")

pblazej:
🍏 

- @hiroshihorie attaching the [airpods-crash.log](https://github.com/user-attachments/files/25386382/airpods-crash.log) that's probably related to VP (aggregate device) and may be related to the above issue ⬆️ I tried to mitigate the exception [here](https://github.com/webrtc-sdk/webrtc/pull/217/commits/484410d16efaae22daddc2e9aa79dbb3761e2a8d), but that's just a symptom. I did not manage to reproduce it tho (sleeps?)
- @cloudwebrtc I did two (suboptimal) changes to fix the av1 resolution, wondering if there's a better idea (upstream bug?)

pblazej (approved):
LGTM I think it's good enough to create .01 release and focus on release testing

- I reverted the global fix for av1 leaving the local one 456316eae443c2f881a5c20a7e8d6b670c143117
- I moved the field trials to the SDK ecfc3dd5e25f7d068115023a881411b7959b2a48

PR #219, opened 2026-02-09
Fix EXC_BAD_ACCESS in inputNode VP property reads on Mac Catalyst background

dfed:
## Summary

Accessing `AVAudioInputNode` voice processing properties (`voiceProcessingEnabled`, etc.) can crash with `EXC_BAD_ACCESS` when the audio hardware is unavailable — primarily when a Mac Catalyst app is in the background.

This PR replaces hardware reads of `inputNode().voiceProcessingEnabled` with cached `EngineState` values in guard conditions throughout `ApplyDeviceEngineState`. When voice processing is disabled, `inputNode` VP properties are never accessed. When VP is enabled, the audio hardware must be active, so the remaining sub-property reads (bypass, AGC, mute, ducking) are safe and preserved for self-healing behavior.

**This is a speculative and untested fix**. It looks good to me, but I'm not terribly familiar with this code, nor do I have a good way to test this locally. Feel free to close if this isn't the direction you'd want to take.

## Crash Details

If helpful, a breakdown of crashes in my app that have hit the issue we're drying to address here:

- **91% of crashes occur in background**, 75% specifically on Mac Catalyst
- Affects apps that call `setVoiceProcessingEnabled(false)`
- `AVAudioInputNode` property access triggers hardware query that crashes when hardware is unavailable
- Sentry: ~300+ crash events across multiple app versions

## Changes

- Use `state.prev.voice_processing_enabled` / `state.next.voice_processing_enabled` (cached) instead of `inputNode().voiceProcessingEnabled` (hardware) in 7 guard conditions
- For the "Configure VP" step, compute `effective_prev_vp` accounting for engine recreate (fresh `AVAudioEngine` defaults VP to disabled)
- VP sub-property reads (`voiceProcessingInputMuted`, `voiceProcessingBypassed`, `voiceProcessingAGCEnabled`) are intentionally kept as hardware reads — they're only reached when VP is enabled (hardware active) and provide self-healing if hardware state drifts

## Ref

- https://github.com/livekit/client-sdk-swift/issues/910

### DIFF
--- a/modules/audio_device/audio_engine_device.mm
+++ b/modules/audio_device/audio_engine_device.mm
@@ -1923,8 +1923,14 @@ int32_t AudioEngineDevice::ApplyDeviceEngineState(EngineStateUpdate state) {
   // --------------------------------------------------------------------------------------------
   // Step: Configure Voice-Processing I/O
   //
+  // Use cached state to avoid accessing inputNode() when voice processing is
+  // disabled – AVAudioInputNode can crash (EXC_BAD_ACCESS) when the audio
+  // hardware is unavailable, e.g. Mac Catalyst in background.
+  // After engine recreate, a fresh AVAudioEngine defaults VP to disabled.
+  bool effective_prev_vp =
+      state.IsEngineRecreateRequired() ? false : state.prev.voice_processing_enabled;
   if (state.next.IsInputEnabled() &&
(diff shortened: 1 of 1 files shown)


pblazej (approved):
lgtm, looking at warnings the aggregate device access must be related to vp

cloudwebrtc (approved):

PR #221, opened 2026-02-20
Audio engine stability improvements

hiroshihorie:
- VP toggle → full engine recreate: Toggling voiceProcessingEnabled now triggers a full engine recreate instead of a restart, ensuring clean audio hardware state and preventing stalled
  capture callbacks after VP reconfiguration.
- Output follows input on all platforms: The output graph now always stays active while input is enabled, not just on macOS or when VP is on. Simplifies logic and prevents capture stalls.
- AudioOutputUnitStop before engine release: Explicitly stops input/output AudioUnits before releasing the engine during VP toggle. Ungated to all platforms.
- Hybrid VP property reads: Guard conditions use cached EngineState values (safe from EXC_BAD_ACCESS when hardware is unavailable, e.g. Mac Catalyst background). The "Configure VP" step
  retains the hardware read for self-healing. Supersedes #219 with a more targeted approach.
- Fix no audio on mute mode switch: Mute mechanisms (VP mute, mixer volume) now declaratively converge to the correct state — each mechanism is always driven to its target regardless of
  mode transitions. Fixes audio loss when switching from InputMixer to VoiceProcessing mode while muted.
- Diagnostic logging: Added detailed [State] / [HW] / [Post] logging to ApplyDeviceEngineState for debugging state transitions.

### DIFF
--- a/modules/audio_device/audio_engine_device.mm
+++ b/modules/audio_device/audio_engine_device.mm
@@ -1791,24 +1791,25 @@ int32_t AudioEngineDevice::ApplyDeviceEngineState(EngineStateUpdate state) {
       (!state.next.IsAnyRunning() || state.IsEngineRestartRequired() ||
        state.DidBeginInterruption() || state.IsEngineRecreateRequired())) {
     LOGI() << "Stopping AVAudioEngine...";
-    RTC_DCHECK(engine_device_ != nil);
 
     if (configuration_observer_ != nullptr) {
       NSNotificationCenter* center = [NSNotificationCenter defaultCenter];
       [center removeObserver:(__bridge_transfer id)configuration_observer_
                         name:AVAudioEngineConfigurationChangeNotification
-                      object:engine_device_];
+                      object:nil];
       configuration_observer_ = nil;
     }
 
-    [engine_device_ stop];
+    if (engine_device_ != nil) {
+      [engine_device_ stop];
 
-    if (observer_ != nullptr) {
-      int32_t result = observer_->OnEngineDidStop(engine_device_, state.next.IsOutputEnabled(),
-                                                  state.next.IsInputEnabled());
-      if (result != 0) {
-        LOGE() << "Call to OnEngineDidStop returned error: " << result;
-        return rollback(result);
+      if (observer_ != nullptr) {
+        int32_t result = observer_->OnEngineDidStop(engine_device_, state.next.IsOutputEnabled(),
+                                                    state.next.IsInputEnabled());
+        if (result != 0) {
+          LOGE() << "Call to OnEngineDidStop returned error: " << result;
+          return rollback(result);
+        }
       }
     }
   }
@@ -1816,7 +1817,8 @@ int32_t AudioEngineDevice::ApplyDeviceEngineState(EngineStateUpdate state) {
   // --------------------------------------------------------------------------------------------
   // Step: Stop playout buffer
   //
-  if (!state.next.IsOutputEnabled() && audio_device_buffer_->IsPlaying()) {
+  if ((!state.next.IsOutputEnabled() || state.IsEngineRecreateRequired()) &&
+      audio_device_buffer_->IsPlaying()) {
     LOGI() << "Stopping Playout buffer...";
     if (engine_device_ != nullptr) {
       // Rendering must be stopped first.
@@ -1828,7 +1830,8 @@ int32_t AudioEngineDevice::ApplyDeviceEngineState(EngineStateUpdate state) {
   // --------------------------------------------------------------------------------------------
   // Step: Stop recording buffer
   //
-  if (!state.next.IsInputEnabled() && audio_device_buffer_->IsRecording()) {
+  if ((!state.next.IsInputEnabled() || state.IsEngineRecreateRequired()) &&
+      audio_device_buffer_->IsRecording()) {
     LOGI() << "Stopping Record buffer...";
     if (engine_device_ != nullptr) {
       // Rendering must be stopped first.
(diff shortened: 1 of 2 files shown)


modules/audio_device/audio_engine_device.mm
```diff
@@ -1842,13 +1925,26 @@ AVAudioVoiceProcessingOtherAudioDuckingLevel ToAVDuckingLevel(
...
         return rollback(result);
       }
     }
+
+    if (state.DidUpdateVoiceProcessingEnabled() && engine_device_ != nil) {
+      AVAudioInputNode* input_node = engine_device_.inputNode;
+      AVAudioOutputNode* output_node = engine_device_.outputNode;
+
+      if (input_node != nil && input_node.audioUnit != nullptr) {
+        AudioOutputUnitStop(input_node.audioUnit);
```
pblazej, line 1937: Trying to understand that, is this stop required only under VP branch?

pblazej:
Haven't noticed anything on iOS/macOS, I'll try to test some different device combinations 🔈

modules/audio_device/audio_engine_device.mm
```diff
@@ -1923,14 +2019,8 @@ AVAudioVoiceProcessingOtherAudioDuckingLevel ToAVDuckingLevel(
...
   //
-  // Use cached state to avoid accessing inputNode() when voice processing is
-  // disabled – AVAudioInputNode can crash (EXC_BAD_ACCESS) when the audio
-  // hardware is unavailable, e.g. Mac Catalyst in background.
-  // After engine recreate, a fresh AVAudioEngine defaults VP to disabled.
-  bool effective_prev_vp =
-      state.IsEngineRecreateRequired() ? false : state.prev.voice_processing_enabled;
   if (state.next.IsInputEnabled() &&
-      effective_prev_vp != state.next.voice_processing_enabled) {
+      inputNode().voiceProcessingEnabled != state.next.voice_processing_enabled) {
```
pblazej: Should it `@catch` as well (reading directly from input node)?

modules/audio_device/audio_engine_device.mm
```diff
@@ -2424,19 +2517,26 @@ AVAudioVoiceProcessingOtherAudioDuckingLevel ToAVDuckingLevel(
   if (state.next.IsAnyEnabled() &&
       (!state.prev.IsAnyEnabled() || state.IsEngineRecreateRequired())) {
     if (state.next.IsInputEnabled()) {
-      uint32_t input_device_id = state.next.input_device_id;
-      if (input_device_id == kAudioObjectUnknown) {
+      uint32_t requested_input_device_id = state.next.input_device_id;
+
+      AudioUnit input_unit = inputNode().audioUnit;
```
pblazej: Nit: could be moved under `else`

modules/audio_device/audio_engine_device.mm
```diff
@@ -1842,13 +1925,26 @@ AVAudioVoiceProcessingOtherAudioDuckingLevel ToAVDuckingLevel(
   //
   if (state.IsEngineRecreateRequired()) {
     LOGI() << "Recreate required, releasing AVAudioEngine...";
-    if (observer_ != nullptr) {
+    if (observer_ != nullptr && engine_device_ != nil) {
       int32_t result = observer_->OnEngineWillRelease(engine_device_);
```
pblazej: Hm, should observers deal with that after or before stop (order)?

modules/audio_device/audio_engine_device.mm
```diff
@@ -1791,32 +1871,34 @@ AVAudioVoiceProcessingOtherAudioDuckingLevel ToAVDuckingLevel(
...
+        }
       }
     }
   }
 
   // --------------------------------------------------------------------------------------------
   // Step: Stop playout buffer
   //
-  if (!state.next.IsOutputEnabled() && audio_device_buffer_->IsPlaying()) {
+  if ((!state.next.IsOutputEnabled() || state.IsEngineRecreateRequired()) &&
```
pblazej, line 1898: Here and below: do we need a rollback action when the stop is triggered by `IsEngineRecreateRequired()` while the buffer is still wanted (additional improvement)?

pblazej:
On macOS while switching devices, external ones fall into this loop with VP enabled, AI log analysis below:

```
  while (!start_result && start_retry_count < kStartEngineMaxRetries)
```

```
  Scenario 1 — VPIO aggregate device retry spiral:
  Engine recreate on input device change → new VPIO can't configure aggregate device because hardware
  is still held by old engine. Retries with incrementing aggregate device IDs, each creating a new
  broken aggregate:

  Timeout waiting for streams
  AUHAL stream format error: mic channel count: 0, numAggInChannels = 0
  couldn't set device 452 on AUHal output bus, err=-10851
  couldn't set device 458 on AUHal output bus, err=-10851
  ...
  couldn't set device 518 on AUHal output bus, err=-10851
  Failure to update stream formats. Max retries exceeded
  AUVoiceProcessor initialize failure. err=-10875
  Engine could not initialize, error = -10875

  Second recreate attempt partially recovers but first render fails:
  [Voice_Processor_Interface_Adapter] failed to run downlink DSP (state fault)

  Scenario 2 — IO thread collision:
  Same recreate path, but old VPIO's IO thread is still alive when the new engine starts:

  Timeout waiting for streams
  HALC_ProxyIOContext::_StartIO(): Start failed - StartAndWaitForState returned error 89
  HALB_IOThread::_Start: there already is a thread
  [Voice_Processor_Interface_Adapter] failed to run downlink DSP (state fault)

  Both point to the same root cause: when VP is active and engine is recreated for a device switch,
  the old VPIO aggregate device / IO thread isn't fully torn down before engine_device_ = nil. The
  explicit AudioOutputUnitStop calls (currently gated on DidUpdateVoiceProcessingEnabled()) would fix
  this if broadened to cover device-switch recreates while VP is enabled.
```

So looks like it may be related to this https://github.com/webrtc-sdk/webrtc/pull/221#discussion_r2847319759 (or the hardcoded `usleep`)

Logs
[device-switch.log](https://github.com/user-attachments/files/25539245/device-switch.log)

pblazej:
We can also think of dispatching the engine recreation internally from this: `public func setVoiceProcessingEnabled(_ enabled: Bool) throws` (still blocks the main thread)

pblazej (commented):
Added a couple of questions, I think iOS is fine, if we can get rid of the blocking loop (or non-blocking loop at least) it would be a good improvement.

hiroshihorie:
@pblazej Thanks for testing, I'll try to repro the macOS issue.

xianshijing-lk:
Do we have a way to repro and problem and verify that this PR fixes the issue ? 

what is the next step for this PR ?

hiroshihorie:
Closing in favor of https://github.com/webrtc-sdk/webrtc/pull/228 .

PR #223, opened 2026-02-24
Allow using custom KDF in `frame_crypto_transformer`

onestacked:
This adds a `key_derivation_function` field to `KeyProviderOptions` to allow the user to define their own KDF.

~I've managed to build webrtc-sdk, but I've not managed to build LiveKit my custom build, so I couldn't test it.~

Fixes: #220

PR #224, opened 2026-03-02
feat: HkdfSha256 key derivation.

cloudwebrtc:


xianshijing-lk:
can you add some tests to make sure things work ?

api/crypto/frame_crypto_transformer.cc
```diff
@@ -234,16 +236,40 @@ uint8_t get_unencrypted_bytes(webrtc::TransformableFrameInterface* frame,
   return unencrypted_bytes;
 }
 
+int DeriveHkdfSha256FromSecret(const std::vector<uint8_t>& secret,
+                               const std::vector<uint8_t>& salt,
+                               unsigned int optional_length_bits,
+                               std::vector<uint8_t>& derived_key) {
+  size_t key_size_bytes = optional_length_bits / 8;
```
xianshijing-lk, line 243: add a assert to make sure optional_length_bits is a multiple of 8

api/crypto/frame_crypto_transformer.cc
```diff
@@ -234,16 +236,40 @@ uint8_t get_unencrypted_bytes(webrtc::TransformableFrameInterface* frame,
...
+  // Use 128 bytes of zeros to padded as info.
+  auto info = std::vector<uint8_t>(128, 0);
+  derived_key.resize(key_size_bytes);
+  if (::HKDF((uint8_t *)derived_key.data(), key_size_bytes, EVP_sha256(), secret.data(),
+             secret.size(), salt.data(), salt.size(), info.data(), info.size()) != 1) {
+    RTC_LOG(LS_ERROR) << "Failed to derive HkdfSha256 key from secret.";
+    return ErrorUnexpected;
+  }
+  RTC_LOG(LS_INFO) << "secret "
```
xianshijing-lk: should you use LS_DEBUG rather than LS_INFO ? btw, is the log necessary ?

xianshijing-lk (approved):
lgtm assuming you will address all the comments and add the unit tests

onestacked:
Is this intended to replace #223?

cloudwebrtc:
Hi @onestacked,  yeah, we considered that besides Rust SDKs, other native SDKs (Swift/Android/Flutter/RN) also need to support HKDF key derivation. Therefore, passing a flag through KeyProviderOptions has the least impact on all platforms.

onestacked:
> Hi @onestacked, yeah, we considered that besides Rust SDKs, other native SDKs (Swift/Android/Flutter/RN) also need to support HKDF key derivation. Therefore, passing a flag through KeyProviderOptions has the least impact on all platforms.

Alright I guess this is reasonable, I mostly took the callback approach to reduce the C++ side changes as much as possible and to be as flexible as possible.
However to avoid losing the ability for custom functions, couldn't you just add a HKDF function in here and from the different SDK just switch between the PBKDF function and HKDF?

onestacked:
I have tried this out with matrix-jukebox (and a locally patched livekit-sdk) and it works with Element Call.

PR #225, opened 2026-03-03
fix: crash on pc.close

pblazej:
WIP https://github.com/livekit/client-sdk-swift/issues/929


pblazej:
Hm looks like the proxy should already handle that... will keep digging.

pblazej:
Closing as looks like the race condition may come from elsewhere and may be fixed by m144

PR #227, opened 2026-03-06
Fix missing RTC_OBJC_TYPE macro on SckHelper

ackanir:
`SckHelper` in `modules/desktop_capture/mac/screen_capturer_sck.mm` is not wrapped with `RTC_OBJC_TYPE()`. When building with `rtc_objc_prefix`, this class is not renamed, causing ObjC runtime collisions if another copy of WebRTC is linked into the same binary.

For context, there were previous omissions covered by these PRs:
- #100
- #171
- #174


cloudwebrtc (approved):
lg!

PR #228, opened 2026-03-17
Audio engine stability fixes 2

hiroshihorie:
Ported from https://github.com/webrtc-sdk/webrtc/pull/221 for m144

- Restore @try/@catch around setVoiceProcessingEnabled (accidentally dropped during port)
- Add @try/@catch guards around all connect:to:format: calls to prevent IsFormatSampleRateAndChannelCountValid crashes (#577)
- Replace hardware VP property reads with state-based vp_was_configured checks (avoids EXC_BAD_ACCESS)
- Stop playout/recording buffers on engine recreate (not just on disable)
- Explicit AudioOutputUnitStop before engine release (unconditional, not VP-gated)
- Log AudioOutputUnitStop results for VPIO teardown observability
- macOS: skip kAudioOutputUnitProperty_CurrentDevice for default routing
- Diagnostic state transition logging

### DIFF
--- a/modules/audio_device/audio_engine_device.h
+++ b/modules/audio_device/audio_engine_device.h
@@ -179,9 +179,8 @@ class AudioEngineDevice : public AudioDeviceModule, public AudioSessionObserver 
     // AUDIO STATE LOGIC
     //
     // Device Mode:
-    // - Output follows input only when voice_processing_enabled=true (for AEC)
+    // - Output follows input to keep AVAudioEngine IO active for capture.
     // - Input respects mute mode restrictions (RestartEngine + input_muted)
-    // - Independent operation when voice processing is disabled
     //
     // Manual Mode:
     // - Bidirectional coupling: if ANY component is enabled/running, BOTH are considered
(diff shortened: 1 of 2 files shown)


pblazej:
Code-wise it looks good to me + comments from the 1st PR 🥇 

I'll try to do some testing soon and maybe let's update for .02

pblazej (approved):
I haven't found anything new, I think it's worth shipping :shipit: 

For some combination of devices macOS may still fall into this loop https://github.com/webrtc-sdk/webrtc/pull/221#issuecomment-3957405276 but haven't tested any actual fixes for that

hiroshihorie:
@pblazej Were you able to repro this loop ?

cloudwebrtc (approved):
LGTM

pblazej:
> Were you able to repro this loop ?

Yes by randomly switching external mics/airpods on macOS, what could be a hint: my mic reports separate devices for rec/playback; not sure if that's the root cause (selecting the "wrong" one).

PR #230, opened 2026-03-31
Fix AudioConverterRef "conv == NULL" causing silent mic capture

brainwith:
## Problem

Microphone capture silently fails with `ca_debug_string: conv == NULL` logged every render cycle (~21ms). Once triggered, all calls in the APP session are affected until restart.

Reproduced on iPhone 7 / iOS 15 but can occur on any device/version since it is undefined behavior.

> **Note:** Older devices (e.g. iPhone 7, 2GB RAM) reproduce more frequently, likely because higher memory pressure leads to more heap reuse, making recycled memory blocks more likely to contain non-zero residual data.

## Root Cause

`converter_ref_` (`AudioConverterRef`, a bare C pointer) in `AudioEngineDevice` is not initialized ([`audio_engine_device.h:550`](https://github.com/webrtc-sdk/webrtc/blob/770e546/modules/audio_device/audio_engine_device.h#L550)). In C++, uninitialized scalar members hold indeterminate values.

The guard at [`audio_engine_device.mm:2304`](https://github.com/webrtc-sdk/webrtc/blob/770e546/modules/audio_device/audio_engine_device.mm#L2304) skips converter creation when `converter_ref_` is non-zero garbage:

```cpp
// converter_ref_ is uninitialized — may hold a non-zero garbage value
if (converter_ref_ == nullptr) {       // ← false when garbage is non-zero
    AudioConverterNew(...);            // ← SKIPPED, converter never created
}
```

All subsequent calls to `AudioConverterConvertComplexBuffer(converter_ref_, ...)` in the sink block use this invalid pointer, causing Core Audio to log `conv == NULL` and discard all captured mic audio.

`AudioEngineDevice` is created once per APP lifecycle, so a bad initial value persists across all calls until the APP is restarted.

## Fix

Add default member initializers to all uninitialized bare pointer members:

- `converter_ref_ = nullptr` (critical — fixes silent capture)
- `observer_ = nullptr`
- `configuration_observer_ = nullptr`

### DIFF
--- a/modules/audio_device/audio_engine_device.h
+++ b/modules/audio_device/audio_engine_device.h
@@ -517,7 +517,7 @@ class AudioEngineDevice : public AudioDeviceModule, public AudioSessionObserver 
   // Set to true after successful call to Init(), false otherwise.
   bool initialized_ RTC_GUARDED_BY(thread_);
 
-  AudioDeviceObserver* observer_ RTC_GUARDED_BY(thread_);
+  AudioDeviceObserver* observer_ RTC_GUARDED_BY(thread_) = nullptr;
 
 #if defined(WEBRTC_IOS)
   // Audio interruption observer instance.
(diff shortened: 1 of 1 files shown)


pblazej:
Is it in m144 only?

hiroshihorie:
lgtm, does this cover all the "uninitialized bare pointer members" ?

brainwith:
> Is it in m144 only?

This issue should exist across all branches. I encountered it in both m125 and m137, and it was finally identified in m144. Based on the root cause, it is likely present in all branches.

brainwith:
> lgtm, does this cover all the "uninitialized bare pointer members" ?

Yes, I audited all member variables in AudioEngineDevice. Only these three are bare C/C++ pointers without initialization. The rest are either Objective-C object pointers (which are zero-initialized under ARC), C++ types with default constructors (e.g., std::unique_ptr, std::vector), or explicitly initialized in the constructor.

brainwith:
Once this issue is triggered, the audio will not recover. The app will remain in a no-audio state even after leaving and rejoining the room. The only way to potentially recover is to restart the app, but even that is not guaranteed, as the uninitialized memory may be hit again.

hiroshihorie (approved):

PR #235, opened 2026-05-04
Expose RtpReceiver::SetJitterBufferMinimumDelay on Obj-C and Java

doracawl:
## Summary

The C++ `RtpReceiverInterface::SetJitterBufferMinimumDelay` has been part of the WebRTC public API for a long time (it backs the W3C `RTCRtpReceiver.jitterBufferTarget` attribute), but neither the Obj-C nor Java wrappers in this fork expose it. Apps building on the prebuilt SDKs therefore have no way to control receive-side jitter buffer minimum delay — i.e. trade latency for smoothness on weak networks — without dropping out to a custom build.

This PR adds:

- **Obj-C**: `-[RTCRtpReceiver setJitterBufferMinimumDelay:]` taking a `nullable NSNumber *`, where `nil` restores default behaviour.
- **Java**: `RtpReceiver.setJitterBufferMinimumDelay(@Nullable Double)`, where `null` restores default behaviour.

Both forward straight to the existing C++ entry point with `std::optional<double>`, mirroring how the browser exposes the property. The public setter rejects non-finite input; the native boundaries (JNI / Obj-C) defensively treat a non-finite value as "clear" — see Design notes.

## Files changed

- `sdk/objc/api/peerconnection/RTCRtpReceiver.h` — protocol declaration
- `sdk/objc/api/peerconnection/RTCRtpReceiver.mm` — implementation
- `sdk/android/api/org/webrtc/RtpReceiver.java` — public Java API + `native` decls
- `sdk/android/src/jni/pc/rtp_receiver.cc` — JNI bindings

Additive only; no behavioural change to existing APIs. The C++ symbol is unchanged and the change is binary-compatible with the current ABI.

## Why this matters downstream

We are building a multi-tenant IoT/video-surveillance mobile app on the prebuilt SDKs from this fork (m144). On weak networks our users see frequent video stutter that a 200–400 ms minimum jitter buffer cleanly absorbs. The C++ method is exactly what we need; without this binding we would have to carry a private SDK build forever. Exposing it once here removes that burden.

## Design notes

1. **Why a nullable `Double` on Java?** It matches this repo's existing convention for optional numeric setters (`EncodedImage.setQp(@Nullable Integer)`, the `@Nullable Integer` parameters on `VideoSource`), keeps the Java surface symmetric with the Obj-C `nullable NSNumber *`, and maps directly onto the C++ `std::optional<double>` (`null`/`nil` → `std::nullopt`).
2. **Why a method on Obj-C, not a property?** The underlying C++ has no getter — a property would lie about being readable.
3. **Why `std::optional`?** m144 already uses `std::optional` throughout `api/rtp_receiver_interface.h`, so we match the existing surface.
4. **Non-finite input.** A NaN reaches a fatal `RTC_CHECK` in `JitterBufferDelay::GetMs` (`saturated_cast<int>` of `delay * 1000`). The Java setter rejects NaN/Inf with `IllegalArgumentException`; the JNI and Obj-C boundaries treat non-finite as `std::nullopt` so a caller reaching native directly cannot crash the engine. Validation uses `Double.isNaN`/`isInfinite` rather than the API 24+ `Double.isFinite`, matching this module's `min_sdk_version` of 23.

## Testing

- All four files compile against the m144 source tree.
- Verified the symbol naming on the JNI side matches the Java native method names (the auto-generated `RtpReceiver_jni.h` derives `JNI_RtpReceiver_<MethodName>` from the Java `nativeXxx` declarations).
- We will be running this against real iOS arm64 devices and Android arm64 devices in our app once a tagged build is available; happy to follow up with stat numbers from `RTCInboundRtpStreamStats.jitterBufferDelay` if useful.


sdk/android/api/org/webrtc/RtpReceiver.java
```diff
@@ -87,6 +87,23 @@ public void setFrameDecryptor(FrameDecryptor frameDecryptor) {
     nativeSetFrameDecryptor(nativeRtpReceiver, frameDecryptor.getNativeFrameDecryptor());
   }
 
+  /**
+   * Sets the minimum jitter buffer delay (in seconds) for this receiver.
+   * The receiver keeps at least this much media buffered before playout,
+   * trading latency for smoothness on weak networks. Wraps the C++
+   * RtpReceiverInterface::SetJitterBufferMinimumDelay.
+   */
```
hiroshihorie, line 99: One small thing on the docs. jitterBufferTarget on the browser side is in milliseconds while this method takes seconds, and I can easily see someone porting browser code getting bitten by a 1000x mistake. A concrete example in the doc should prevent that, and the internal clamp is worth mentioning too since the native side silently caps at 10 seconds.

```suggestion
  /**
   * Sets the minimum jitter buffer delay in seconds for this receiver.
   * The receiver keeps at least this much media buffered before playout,
   * trading latency for smoothness on weak networks. Note the unit: the
   * equivalent browser attribute, RTCRtpReceiver.jitterBufferTarget, is
   * expressed in milliseconds, so a browser value of 500 corresponds to
   * 0.5 here. Values are clamped to [0, 10] seconds internally. Wraps the
   * C++ RtpReceiverInterface::SetJitterBufferMinimumDelay.
   */
```

sdk/objc/api/peerconnection/RTCRtpReceiver.h
```diff
@@ -93,6 +93,14 @@ the current RTCRtpReceiver in the last ten seconds.
 /** The delegate for this RtpReceiver. */
 @property(nonatomic, weak) id<RTC_OBJC_TYPE(RTCRtpReceiverDelegate)> delegate;
 
+/** Sets the minimum jitter buffer delay (in seconds) for this receiver.
+ *  Pass nil to restore the default behaviour. The receiver keeps at least
+ *  this much media buffered before playout, trading latency for smoothness
+ *  on weak networks. Mirrors RTCRtpReceiver.jitterBufferTarget in browser
+ *  WebRTC; wraps the C++ RtpReceiverInterface::SetJitterBufferMinimumDelay.
+ */
```
hiroshihorie: Same units note for the ObjC side. Mentioning jitterBufferTarget is great for discoverability, just worth being explicit that the units differ.

```suggestion
/** Sets the minimum jitter buffer delay in seconds for this receiver.
 *  Pass nil to restore the default behaviour. The receiver keeps at least
 *  this much media buffered before playout, trading latency for smoothness
 *  on weak networks. Note the unit: the equivalent browser attribute,
 *  RTCRtpReceiver.jitterBufferTarget, is expressed in milliseconds, so a
 *  browser value of 500 corresponds to 0.5 here. Values are clamped to
 *  [0, 10] seconds internally. Wraps the C++
 *  RtpReceiverInterface::SetJitterBufferMinimumDelay.
 */
```

hiroshihorie (commented):
Thanks for the contribution! The implementation looks great, both sides forward the optional exactly like the C++ API expects and the Java setter/clear split keeps the hot path allocation free, nice touch. The guard and naming also match the existing patterns in these wrappers.

I have one small suggestion about the docs, see inline. It is about units, the only thing I would really want fixed before merge.

doracawl:
Thanks for the review and the units catch — the ms-vs-seconds gap against `jitterBufferTarget` is an easy trap, so I've folded your wording (including the [0, 10] s clamp note) into the doc comments on both sides.

While there, I aligned the Java surface with this repo's own convention for optional numeric setters — `EncodedImage.setQp(@Nullable Integer)` and the `@Nullable Integer` parameters on `VideoSource` — by collapsing the previous `set(double)` + `clear()` pair into a single `setJitterBufferMinimumDelay(@Nullable Double)` where `null` restores the default. That also keeps it symmetric with the Obj-C `nullable NSNumber *`.

I also adopted the Copilot note on non-finite input — it's a genuine crash path (`saturated_cast<int>` in `JitterBufferDelay::GetMs` fatally `RTC_CHECK`s on NaN). The Java setter now rejects NaN/Inf with `IllegalArgumentException`, and the JNI / Obj-C boundaries treat non-finite as `nullopt` for callers reaching native directly. The validation deliberately uses `Double.isNaN`/`isInfinite` rather than `Double.isFinite`, since the latter is API 24+ and this module's `min_sdk_version` is 23. PR description updated to match — happy to keep the explicit `clear()` instead if you'd prefer.

hiroshihorie (approved):
Re-reviewed the latest revision. The units docs are applied on both sides and match the actual behavior in pc/jitter_buffer_delay.cc, including the [0, 10] s clamp. The NaN guard is correct. saturated_cast<int> hits RTC_CHECK_NOTREACHED() on NaN, and the layering (Java throws, JNI and Obj-C map non-finite to nullopt) is sensible. Threading is safe since SetJitterBufferMinimumDelay is proxied to the worker thread, and the JNI naming matches the generated header convention.

One optional nit, not blocking: Java throws on non-finite input while Obj-C silently resets to default, and the Obj-C header doc does not mention that. A one-line addition like "non-finite values are treated as nil" would make it fully self-documenting. Fine to fix in a follow-up or leave as is.

Thanks for addressing the feedback.

doracawl:
Added the one-liner to the Obj-C header doc — non-finite values are now documented as being treated as nil (88c4bdb). Thanks for the review!

PR #238, opened 2026-05-11
Fix missing RTC_OBJC_TYPE macro on SckHelper

ackanir:
Porting the #227 patch (already merged) from m137 to m144.


cloudwebrtc (approved):
lg!

PR #243, opened 2026-05-14
Fix H.264 h/w acceleration

pblazej:
Should allow h/w acceleration even if the base profile is negotiated:

```
(RTCVideoEncoderH264.mm:407): Using profile H264_Baseline_5_0
(RTCVideoEncoderH264.mm:747): H264: skipping EnableLowLatencyRateControl (non-High profile family).
(RTCVideoEncoderH264.mm:774): Compression session created with hw accl enabled
```

Follow-up of https://github.com/livekit/client-sdk-swift/issues/1002 and maybe https://github.com/livekit/client-sdk-swift/issues/1021

Edit: Unfortunately this problem also occurs for H.265, but I haven't found any reasonable solution/workaround, so keeping it small...

### DIFF
--- a/sdk/objc/components/video_codec/RTCVideoEncoderH264.mm
+++ b/sdk/objc/components/video_codec/RTCVideoEncoderH264.mm
@@ -359,6 +359,19 @@ NSUInteger GetMaxSampleRate(
       return 0;
   }
 }
+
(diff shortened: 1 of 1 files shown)


pblazej:
Build: https://github.com/webrtc-sdk/webrtc-build/actions/runs/26504879401

hiroshihorie (approved):

PR #246, opened 2026-05-21
No stop on mute for linux

hiroshihorie:


cloudwebrtc (approved):

hiroshihorie:
Closing this for now since, I think Stop-on-mute should be default behavior . Platforms that properly implement the ADM's Mute API can opt out individually, like iOS/macOS ADM.

PR #248, opened 2026-05-28
Fix H.265 h/w acceleration

pblazej:
Follow-up of #243 looks like we need to disable it unconditionally to get:

```
(RTCVideoEncoderH265.mm:433): Compression session created with hw accl enabled
```


hiroshihorie (approved):
✅ Hm... interesting ...

PR #249, opened 2026-06-04
Fix macOS audio build warnings

hiroshihorie:
## Summary
- Replace deprecated CoreAudio `kAudioObjectPropertyElementMaster` uses with `kAudioObjectPropertyElementMain`.
- Avoid repeated ObjC++ `dynamic_cast` checks in `RTCAudioDeviceModule` by carrying the selected ADM type into the wrapper.
- Keep AudioEngine-only wrapper APIs gated behind `RTCAudioDeviceModuleTypeAudioEngine`.

## Details
The public factory path already creates `AudioEngineDevice` when `audioDeviceModuleType` is `AudioEngine`. This PR preserves that behavior and uses the module type as the wrapper discriminator for AudioEngine-only calls.

For native/custom ADM injection, callers should keep the provided `audioDeviceModuleType` consistent with the actual ADM instance. A custom non-AudioEngine ADM should not be tagged as `AudioEngine`.

## Validation
- `git diff --check m144_release...HEAD`
- `autoninja -C out/codex-mac sdk:peerconnectionfactory_base_objc`

### DIFF
--- a/modules/audio_device/audio_engine_device.mm
+++ b/modules/audio_device/audio_engine_device.mm
@@ -266,7 +266,7 @@ int32_t AudioEngineDevice::Init() {
   // main thread to issue notifications.
   AudioObjectPropertyAddress propertyAddress = {kAudioHardwarePropertyRunLoop,
                                                 kAudioObjectPropertyScopeGlobal,
-                                                kAudioObjectPropertyElementMaster};
+                                                kAudioObjectPropertyElementMain};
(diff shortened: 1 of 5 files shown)


hiroshihorie:
Build https://github.com/webrtc-sdk/webrtc-build/actions/runs/26940364097

cloudwebrtc (approved):

PR #250, opened 2026-06-04
Reduce frame cryptor log noise

hiroshihorie:
## Summary
- Stop logging disabled frame cryptor pass-through as a warning on every frame.
- Keep empty-frame logs at verbose level.
- Treat `enabled=false` as explicit pass-through for both encrypt and decrypt paths.

## Behavior
Previously `data_in.size() == 0` and `enabled_cryption == false` shared the same warning and discard handling. This split keeps discard behavior for empty frames, while disabled cryptors now pass frames through without warning spam.

## Validation
- `git diff --check m144_release...HEAD`
- `autoninja -C out/codex-api-test api/crypto:frame_crypto_transformer`

## Notes
- Full `rtc_unittests` is currently blocked on `m144_release` by unrelated macOS `kAudioObjectPropertyElementMaster` deprecation warnings.
- Focused `api/crypto:crypto_unittests` is blocked by an existing `system_wrappers/include/sleep.h` include-path issue in that target.

### DIFF
--- a/api/crypto/frame_crypto_transformer.cc
+++ b/api/crypto/frame_crypto_transformer.cc
@@ -469,12 +469,16 @@ void FrameCryptorTransformer::encryptFrame(
   }
 
   webrtc::ArrayView<const uint8_t> data_in = frame->GetData();
-  if (data_in.size() == 0 || !enabled_cryption) {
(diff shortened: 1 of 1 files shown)


hiroshihorie:
Build https://github.com/webrtc-sdk/webrtc-build/actions/runs/26941032477

cloudwebrtc (approved):

PR #253, opened 2026-06-08
Allow keyring size 256 and bounds-guard GetKeySet

### LINKED ISSUE #1030: E2EE KeyProvider crashes with keyRingSize 256 and key index 255
**Describe the bug**

`BaseKeyProvider` crashes the process when E2EE `KeyProviderOptions` is configured with `keyRingSize: 256` and a key is set/exported at index `255`.

This looks like a native `LiveKitWebRTC` key ring bounds issue. The Swift API accepts `keyRingSize: 256`, but the process aborts instead of either supporting key index `255` or rejecting the configuration/index gracefully.

**SDK Version**

Reproduced with:

- LiveKit Swift SDK `2.14.1`
  - `LiveKitWebRTC 144.7559.06`
- Current `client-sdk-swift` `main`
  - `LiveKitWebRTC 144.7559.07`

**iOS/macOS Version**

Reproduced on macOS 15.7.3 (24G419).

**Xcode Version**

Reproduced with Xcode 26.3 (17C529), Swift 6.2.4.

**Steps to Reproduce**

Run this minimal Swift code:

```swift
import LiveKit

let options = KeyProviderOptions(
    sharedKey: true,
    keyRingSize: 256
)

let keyProvider = BaseKeyProvider(options: options)

keyProvider.setKey(key: "test-key", index: 255)
_ = keyProvider.exportKey(index: 255)
```

Control case:

```swift
let options = KeyProviderOptions(
    sharedKey: true,
    keyRingSize: 256
)

let keyProvider = BaseKeyProvider(options: options)

keyProvider.setKey(key: "test-key", index: 254)
_ = keyProvider.exportKey(index: 254)
```

**Examples**

The problem is reproducible with:

- [ ] [client-example-swift](https://github.com/livekit/client-example-swift)
- [ ] [client-example-collection-swift](https://github.com/livekit/client-example-collection-swift)

The issue does not require an example app, a LiveKit server, or a room connection to reproduce. It reproduces with the minimal SwiftPM snippet above.

**Expected behavior**

Either:

- `keyRingSize: 256` should support valid key indices `0...255`, or
- the SDK should reject `keyRingSize: 256` / `index: 255` gracefully without aborting the process.

A public Swift API call should not terminate the process due to a native bounds assertion.

**Screenshots**

N/A

**Logs**

Crash output:

```text
libc++ Hardening assertion __n < size() failed: vector[] index out of bounds
```

Additional verification:

- `keyRingSize: 256`, `index: 254` works.
- `keyRingSize: 256`, `index: 255` crashes.
- The crash reproduces in LiveKit Swift SDK `2.14.1` with `LiveKitWebRTC 144.7559.06`.
- The crash also reproduces on current `client-sdk-swift` `main` with `LiveKitWebRTC 144.7559.07`.

**Context**

This matters for E2EE integrations that use one-byte key indices and may need to use key index `255`.

pblazej:
Fixes livekit/client-sdk-swift#1030 (companion test reproducer: livekit/client-sdk-swift#1032).

The Swift/Android/Flutter/Rust SDKs abort the process with `libc++ Hardening assertion __n < size() failed: vector[] index out of bounds` when a caller looks up key index 255 in a key ring configured for size 256. The root cause is `MAX_KEYRING_SIZE = 255` silently clamping a requested 256 to a 255-entry vector, combined with `GetKeySet()` indexing the vector with no bounds check. The JS reference implementation already validates `keyringSize` as `1..=256` ([client-sdk-js ParticipantKeyHandler.ts L43–L46](https://github.com/livekit/client-sdk-js/blob/main/src/e2ee/worker/ParticipantKeyHandler.ts#L43-L46)) and its lookup returns `undefined` rather than aborting on out-of-range reads ([L210](https://github.com/livekit/client-sdk-js/blob/main/src/e2ee/worker/ParticipantKeyHandler.ts#L210)). This PR bumps `MAX_KEYRING_SIZE` to 256 — matching both the documented contract ("Keyring size needs to be between 1 and 256") and the JS reference — and makes `GetKeySet()` return `nullptr` for out-of-range indices. Every existing caller (`ExportSharedKey`, `ExportKey`, `RatchetKey`, the decrypt path) already null-checks the result, so this turns the abort into the same observable failure JS produces.

### DIFF
--- a/api/crypto/frame_crypto_transformer.h
+++ b/api/crypto/frame_crypto_transformer.h
@@ -32,7 +32,7 @@
 namespace webrtc {
 
 const size_t DEFAULT_KEYRING_SIZE = 16;
-const size_t MAX_KEYRING_SIZE = 255;
+const size_t MAX_KEYRING_SIZE = 256;
 
 class ParticipantKeyHandler;
 
@@ -153,7 +153,11 @@ class ParticipantKeyHandler : public webrtc::RefCountInterface {
 
   virtual webrtc::scoped_refptr<KeySet> GetKeySet(int key_index) {
     webrtc::MutexLock lock(&mutex_);
-    return crypto_key_ring_[key_index != -1 ? key_index : current_key_index_];
+    int idx = key_index != -1 ? key_index : current_key_index_;
+    if (idx < 0 || static_cast<size_t>(idx) >= crypto_key_ring_.size()) {
+      return nullptr;
+    }
+    return crypto_key_ring_[idx];
   }
 
   virtual void SetKey(std::vector<uint8_t> password, int key_index) {
--- a/api/crypto/frame_crypto_transformer_unittest.cc
+++ b/api/crypto/frame_crypto_transformer_unittest.cc
@@ -241,4 +241,36 @@ TEST(KeyProvider, KeyDerivationAlgorithm) {
   EXPECT_NE(encrypted_data.value()->iv, encrypted_data2.value()->iv);
 }
 
+TEST(KeyProvider, KeyRingSizeMaxAcceptsIndex255) {
+  auto key_options = KeyProviderOptions();
+  key_options.shared_key = true;
+  key_options.key_ring_size = static_cast<int>(MAX_KEYRING_SIZE);
+  auto key_provider =
+      webrtc::make_ref_counted<DefaultKeyProviderImpl>(key_options);
+
+  const std::vector<uint8_t> key{0xAA, 0xBB, 0xCC, 0xDD};
+  EXPECT_TRUE(key_provider->SetSharedKey(255, key));
+  EXPECT_EQ(key_provider->ExportSharedKey(255), key);
+}
+
+TEST(KeyProvider, GetKeySetReturnsNullptrForOutOfRange) {
+  auto key_options = KeyProviderOptions();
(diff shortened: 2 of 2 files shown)


api/crypto/frame_crypto_transformer.h
```diff
@@ -153,7 +153,11 @@ class ParticipantKeyHandler : public webrtc::RefCountInterface {
 
   virtual webrtc::scoped_refptr<KeySet> GetKeySet(int key_index) {
     webrtc::MutexLock lock(&mutex_);
-    return crypto_key_ring_[key_index != -1 ? key_index : current_key_index_];
+    int idx = key_index != -1 ? key_index : current_key_index_;
+    if (idx < 0 || static_cast<size_t>(idx) >= crypto_key_ring_.size()) {
```
pblazej, line 157: Defensive only — the off-by-one bump above alone fixes #1030; this mirrors [JS `cryptoKeyRing[oob]` returning `undefined`](https://github.com/livekit/client-sdk-js/blob/main/src/e2ee/worker/ParticipantKeyHandler.ts#L210) and closes the same class of crash at any further out-of-range index.

cloudwebrtc (approved):

PR #204, opened 2025-11-01
Removing mic perms check logic

hiroshihorie:
Perhaps we should move this logic to the SDK side with extra logic.

PR #265, opened 2026-07-08
Remove blocking mic permission request from AudioEngine pre-enable check (m144)

hiroshihorie:
Forward-port of #204 (originally targeting `m137_release`) to `m144_release`, which is the line currently shipped by the LiveKit Swift SDK (`main` pins `144.7559.10`). The blocking call was never ported forward, so it still lives on `m144_release`.

## Problem

`EnsureMicrophonePermissionSync()` blocks the WebRTC worker thread on a semaphore (`dispatch_semaphore_wait(…, DISPATCH_TIME_FOREVER)`) until the user answers the mic permission dialog. When the app is woken in the background (e.g. an incoming CallKit call), no dialog can appear, so `setEngineAvailability` hangs indefinitely.

Ref: livekit/client-sdk-swift#815

## Change

Replace the blocking request with a passive `IsMicrophonePermissionAuthorized()` check in the AudioEngine pre-enable path. When the mic is not authorized it returns `kAudioEngineErrorInsufficientDevicePermission` instead of blocking. The (dead) `IsMicrophonePermissionGranted()` is renamed into that check; it had no other callers.

Only the `.notDetermined` case changes behavior: previously it prompted and blocked the worker thread; now it returns an error immediately. `authorized` and `denied`/`restricted` are unchanged.

### DIFF
--- a/modules/audio_device/audio_engine_device.h
+++ b/modules/audio_device/audio_engine_device.h
@@ -512,8 +512,7 @@ class AudioEngineDevice : public AudioDeviceModule, public AudioSessionObserver 
   std::vector<std::string> input_device_labels_;
 #endif
 
-  bool IsMicrophonePermissionGranted();
-  bool EnsureMicrophonePermissionSync();
+  bool IsMicrophonePermissionAuthorized();
 
 #if !TARGET_OS_OSX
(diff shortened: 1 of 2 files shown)


pblazej (approved):
LGTM

PR #266, opened 2026-07-08
android: fix AudioRecord released under a live reader when the stop join times out

KD6763:
## Problem

Production crash on Android (Samsung SM-G766U1, Android 16, LiveKit consumer of `io.github.webrtc-sdk:android-prefixed:144.7559.05`) during rapid capture stop/start cycling while the capture HAL was under pressure (Bluetooth SCO route churn plus a second concurrent capture stack):

```
Abort message: 'releaseBuffer: mUnreleased out of range,
!(stepCount:480 <= mUnreleased:0 <= mFrameCount:3840), BufferSizeInFrames:3840'
  android::ClientProxy::releaseBuffer
  android::AudioRecord::read
  android.media.AudioRecord.read
  livekit.org.webrtc.audio.WebRtcAudioRecord$AudioRecordThread.run
```

`stopRecordingIfNeededImpl()` joins the reader thread while holding `audioRecordStateLock` and never stops the `AudioRecord` first:

1. The reader takes that lock at the top of every loop iteration, so a stop racing a loop boundary deadlocks against the join until its 2s timeout (thread waits for the lock, join waits for the thread). Device logs showed this ~2s stall on every stop cycle under churn.
2. A reader stuck inside a blocking `read()` (stalled HAL) has nothing to unblock it, so the join times out structurally.
3. On timeout the record is released anyway — **under a live reader potentially still inside `read()`** — corrupting the platform client proxy (the abort above). The leaked thread then re-snapshots the shared `audioRecord` field each iteration and its exit path stops whatever the field holds, so once a new start installs a successor record, two readers interleave on one record. The crash tombstone showed two of this class's reader threads alive, started ~400ms apart.

## Fix

One commit per hazard:

1. **Stop and detach the record before joining, join outside the lock** — a blocked `read()` returns once the record leaves the recording state, so the join actually succeeds; detaching under the lock means concurrent init/start builds fresh state.
2. **Never release under a live reader** — on join timeout, release ownership transfers to the reader via an atomic handoff (`orphanedRecord`/`cleanupDone`); exactly one side releases in every interleaving.
3. **Reader-side guards** — the reader re-checks `keepAlive` after the loop-top snapshot and around re-init/start (all under the lock), tracks the record it actually read (`activeRecord`) and stops only that on exit, takes the startup assert under the lock, and no longer misreports a shutdown-stopped record as a start failure (which permanently disabled `useAudioRecord`).

## Verification

Stalled-read reproduction (JVM harness emulating a blocked `read()` that only returns when the record leaves the recording state), unpatched vs patched:

- Unpatched: `release()` lands while a read is in flight — the abort's exact precondition — in 5/5 stop/restart cycles; every stop burns the full 2s join timeout.
- Patched: every stop completes in milliseconds, reads observe a clean `stop()`, no release-under-read, no leaked reader threads.

On-device stress (SM-G766U1, 400 init/start/stop cycles with a concurrent prewarm/init thread): unpatched **346/400** stops hit the 2s join timeout with leaked reader threads; patched **0/400**, max stop 189ms, no crash.

The harness is Robolectric-based (shadowed `AudioRecord` source) so it doesn't fit this tree's GN build; happy to contribute it in whatever form fits your test setup.

## Notes

- Bounded trade-offs: if a join still times out (pathologically stuck read), the orphaned record is released when the reader finally exits — if it never exits the record stays held, which is strictly better than the abort; in that same path the `AUDIO_RECORD_STOP` state callback can be delivered after a successor session's `AUDIO_RECORD_START` (it reflects the actual late stop).
- `m146_release` carries the same stop-path code; happy to open a matching PR there.

### DIFF
--- a/sdk/android/src/java/org/webrtc/audio/WebRtcAudioRecord.java
+++ b/sdk/android/src/java/org/webrtc/audio/WebRtcAudioRecord.java
@@ -663,11 +663,11 @@ class WebRtcAudioRecord {
     Logging.d(TAG, "stopRecordingIfNeeded");
     synchronized(audioRecordStateLock) {
       clientCalledStartRecording.set(false);
-      if(audioThread != null) {
-        return stopRecordingIfNeededImpl();
+      if (audioThread == null) {
+        return true;
       }
     }
-    return true;
+    return stopRecordingIfNeededImpl();
   }
 
   @CalledByNative
@@ -676,19 +676,25 @@ class WebRtcAudioRecord {
     synchronized(audioRecordStateLock) {
       nativeCalledStartRecording.set(false);
       nativeCalledInitRecording.set(false);
-      return stopRecordingIfNeededImpl();
     }
+    return stopRecordingIfNeededImpl();
   }
 
   private boolean stopRecordingIfNeededImpl() {
+    final AudioRecordThread stoppingThread;
+    final @Nullable AudioRecord stoppingRecord;
     synchronized(audioRecordStateLock) {
       if(clientCalledStartRecording.get() || nativeCalledStartRecording.get()) {
         // Someone has still requested recording, ignore stop request.
         return true;
       }
 
+      if (audioThread == null) {
+        // Already stopped by a concurrent caller.
+        return true;
+      }
+
       Logging.d(TAG, "stopping recording");
-      assertTrue(audioThread != null);
       if (future != null) {
         if (!future.isDone()) {
           // Might be needed if the client calls startRecording(), stopRecording() back-to-back.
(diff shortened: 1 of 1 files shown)


davidliu:
@KD6763  Hey, thanks for the PR. Just making sure, this is the same changes as #264?

KD6763:
@davidliu, yes same changes as #264

davidliu (approved):
LGTM!

If you'd be willing to add the test harness in a separate PR, that'd be helpful too. The tests at [sdk/android/tests/src/org/webrtc/audio](https://github.com/webrtc-sdk/webrtc/tree/m144_release/sdk/android/tests/src/org/webrtc/audio) are Robolectric capable (uses `AndroidJUnit4` as its runner).

You'd just need to add the path to the file in the BUILD.gn here:
https://github.com/webrtc-sdk/webrtc/blob/08b9408717876048bca582530ba6f3e732a02479/sdk/android/BUILD.gn#L1914

KD6763:
Thanks! Will do - I'll port the harness to a Robolectric test under sdk/android/tests/src/org/webrtc/audio and put up a separate PR for it.

Also do you want a matching PR for m146_release (it has the same stop path code), or would you rather pick it across on your side?

davidliu:
> Also do you want a matching PR for m146_release (it has the same stop path code), or would you rather pick it across on your side?

We'll handle it on our side, we're not using m146_release actually, we're in the midst of an upgrade to m150.

PR #267, opened 2026-07-09
Fix H.265 decoder crushing 10-bit output and discarding bitstream colorimetry (HDR10 renders washed out)

oliversluke:
## Problem

`RTCVideoDecoderH265` breaks HEVC Main10 / HDR10 reception in two ways:

1. The VTDecompressionSession's destination image buffer attributes force `kCVPixelFormatType_420YpCbCr8BiPlanarFullRange`, so 10-bit output is crushed to 8 bits.
2. `overrideColorSpaceAttachments()` runs unconditionally on every decoded frame, replacing genuine colorimetry from the bitstream VUI (e.g. SMPTE ST 2084 PQ transfer + BT.2020 primaries) with BT.709/sRGB.

Together, any HDR10 stream (HEVC Main10 with PQ/BT.2020 VUI) decodes as washed-out 8-bit SDR: bit depth is lost at the decoder boundary, and downstream renderers see frames mistagged as sRGB/709.

## Fix

- Request a 10-bit biplanar output format (`kCVPixelFormatType_420YpCbCr10BiPlanarVideoRange`) when the hvcC decoder configuration record signals `bit_depth_luma_minus8 > 0`; 8-bit streams keep NV12 exactly as before.
- Apply the BT.709/sRGB attachments only as a *fallback* when the format description carries no colour information from the bitstream — preserving the original deterministic-colour behaviour for VUI-less streams, while leaving real VUI colorimetry (which VideoToolbox propagates from the format description onto output buffers) untouched.

## Compatibility

- 8-bit streams without VUI colour info: byte-for-byte unchanged (NV12 output + BT.709/sRGB fallback tagging).
- 8-bit streams *with* VUI colour info: now correctly tagged instead of overridden.
- Main10 streams: 10-bit `x420` output with true colorimetry. Consumers rendering via `AVSampleBufferDisplayLayer` display HDR correctly; `RTCMTLVideoView`'s shaders may need 10-bit support to benefit, but Main10 content was already colour-broken before this change. Happy to gate the pixel-format change further if you'd prefer.

## Testing

The behaviour change was verified on-device in a receive-only tvOS WebRTC client with 4K HEVC streams: Main10/HDR10 streams now decode as 10-bit with SMPTE ST 2084 / BT.2020 attachments and render correctly, and 8-bit streams are unaffected. I could not build the full webrtc tree locally, so this patch relies on CI for compile verification — happy to iterate on review feedback.

### DIFF
--- a/sdk/objc/components/video_codec/RTCVideoDecoderH265.mm
+++ b/sdk/objc/components/video_codec/RTCVideoDecoderH265.mm
@@ -222,6 +222,16 @@ void h265DecompressionOutputCallback(void *decoderRef, void *params, OSStatus st
   webrtc::RTCVideoFrameReorderQueue _reorderQueue;
 }
 
+static BOOL gPreferHighBitDepthOutput = NO;
+
++ (BOOL)preferHighBitDepthOutput {
+  return gPreferHighBitDepthOutput;
+}
+
++ (void)setPreferHighBitDepthOutput:(BOOL)preferHighBitDepthOutput {
+  gPreferHighBitDepthOutput = preferHighBitDepthOutput;
+}
+
 - (instancetype)init {
   self = [super init];
   if (self) {
@@ -440,9 +450,12 @@ CMSampleBufferRef H265BufferToCMSampleBuffer(const uint8_t *buffer, size_t buffe
       kCVPixelBufferIOSurfacePropertiesKey, kCVPixelBufferPixelFormatTypeKey};
   CFDictionaryRef ioSurfaceValue = CreateCFTypeDictionary(nullptr, nullptr, 0);
   // Forcing NV12 would crush high bit depth output (e.g. HEVC Main10) to
-  // 8 bits. Request a 10-bit biplanar format for such streams; 8-bit streams
-  // keep NV12 so existing consumers are unaffected.
-  int64_t pixelFormatType = [self isHighBitDepthFormat]
+  // 8 bits. Request a 10-bit biplanar format for such streams, but only when
+  // the app has opted in via preferHighBitDepthOutput — downstream consumers
+  // that assume 8-bit NV12 (RTCMTLVideoView, RTCCVPixelBuffer's I420
+  // conversion) do not handle 10-bit output. 8-bit streams keep NV12 either
+  // way.
+  int64_t pixelFormatType = (gPreferHighBitDepthOutput && [self isHighBitDepthFormat])
       ? kCVPixelFormatType_420YpCbCr10BiPlanarVideoRange
       : kCVPixelFormatType_420YpCbCr8BiPlanarFullRange;
   CFNumberRef pixelFormat = CFNumberCreate(nullptr, kCFNumberLongType, &pixelFormatType);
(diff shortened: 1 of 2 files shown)


sdk/objc/components/video_codec/RTCVideoDecoderH265.mm
```diff
@@ -183,7 +184,14 @@ void h265DecompressionOutputCallback(void *decoderRef, void *params, OSStatus st
     return;
   }
 
-  overrideColorSpaceAttachments(imageBuffer);
+  // Only guess colour attachments when the bitstream signalled none. Streams
+  // that carry VUI colour information (e.g. HDR10: PQ transfer + BT.2020
+  // primaries) already have correct attachments propagated by VideoToolbox
+  // from the format description; overriding them mistags the frames as
+  // BT.709/sRGB and HDR content renders washed out.
+  if (![decoder bitstreamCarriesColorInfo]) {
```
pblazej: This runs on VideoToolbox's async callback thread and reads `_videoFormat`, which `decodeData:` releases (via `setVideoFormat:`) *before* `destroyDecompressionSession` waits for in-flight frames — a use-after-free window on every mid-stream format change; consider capturing this as a bool in `RTCH265FrameDecodeParams` at decode time instead.

pblazej:
Thanks — verified this end-to-end with a Main10/PQ stream and the colorimetry + 10-bit output work as described. One question on rollout: since x420 output breaks consumers that assume 8-bit NV12 (`RTCMTLVideoView`'s R8/RG8 shaders render black, `RTCCVPixelBuffer toI420`/`cropAndScale` hit `RTC_DCHECK_NOTREACHED`), what's your vision for hardening this — you mentioned gating the pixel-format change further; would you be open to an explicit opt-in (e.g. a class property or field trial) so Main10 keeps today's NV12 downconversion by default, and/or a lossy 10→8 fallback in `toI420`? Happy to help with either.

sdk/objc/components/video_codec/RTCVideoDecoderH265.mm
oliversluke: Good catch — the callback could indeed read `_videoFormat` in the window where `decodeData:` had already released the old format while the previous session's frames were still in flight (and even after the swap, in-flight frames would be judged against the wrong format's colour info). Fixed as you suggested in d4af43d: the flag is now captured in `RTCH265FrameDecodeParams` on the decode thread, so the callback no longer touches decoder state and each frame is judged against the format it was actually decoded with.

oliversluke:
Thanks for verifying end-to-end! Agreed on the rollout concern — ff6af96 keeps 8-bit NV12 downconversion as the default and gates the x420 output behind an opt-in class property, `RTCVideoDecoderH265.preferHighBitDepthOutput`. The colorimetry fix stays unconditional since it only changes streams whose VUI already carries colour info.

I went with a class property over a field trial because it's the easiest to reach for apps consuming the prebuilt xcframework — happy to switch if you'd prefer the field-trial route. For the lossy 10→8 fallback in `toI420` (and 10-bit-aware shaders in `RTCMTLVideoView`), I'd suggest a follow-up PR so this one stays reviewable — glad to take that on or collaborate on it.

PR #206, opened 2025-11-18
Change default to stop recording when muted

hiroshihorie:

PR #269, opened 2026-07-09
Android: implement ADM SetMicrophoneMute and allow opting out of stop-on-mute

max-buster:
## Summary

Since #206, `WebRtcVoiceSendChannel::MuteStream` stops the ADM's recording when all send streams are muted, unless the ADM overrides `IsStopOnMuteModeEnabled()` to false. The iOS/macOS `AudioEngineDevice` opts out and mutes internally while keeping capture alive; the Android ADM has no override, so muting an audio track now fully releases `AudioRecord`.

Per the guidance on #206 — platforms that properly implement the ADM's Mute API can opt out individually — this PR does exactly that for Android, keeping stop-on-mute as the default.

## Problem

For push-to-talk style apps, re-enabling the mic is a cold start: a new `AudioRecord`, AGC reconvergence, and on USB audio devices a full interface renegotiation (alternate setting selection, bandwidth reservation, sample rate setup). Speech during that window is lost, heard as a clipped first word on every transmission. Before #206, muting kept capture warm on Android. We bisected across LiveKit Android v2.23.0 (webrtc 137.7151.04) / v2.23.1 (137.7151.05) and confirmed the boundary matches this change.

Android already had a working warm-mute implementation — `WebRtcAudioRecord.setMicrophoneMute()` zero-fills the capture buffer on the record thread — but it was only reachable from Java. The C++ `AndroidAudioDeviceModule::SetMicrophoneMute()` returned -1, so the non-stop mute path in `MuteStream` was a no-op on Android, and there was no way to opt out of stop-on-mute.

## Changes

1. **Wire `SetMicrophoneMute` through to `WebRtcAudioRecord`.** New `AudioInput::SetMicrophoneMute` (default -1 for other inputs, preserving behavior) and `AudioRecordJni::SetMicrophoneMute` bridge the C++ ADM to the existing Java zero-fill logic. `AndroidAudioDeviceModule::SetMicrophoneMute` now returns 0 and delegates; `MicrophoneMute()` reports the cached state. This is the same "mute inside the engine, capture stays alive" semantics as iOS `AudioEngineDevice`.
2. **Add builder opt-out.** `JavaAudioDeviceModule.Builder.setStopRecordingOnMute(boolean)` — default `true`, current behavior unchanged. Plumbed through `nativeCreateAudioDeviceModule` into `AndroidAudioDeviceModule`, which overrides `IsStopOnMuteModeEnabled()`. The `CreateAudioDeviceModuleFromInputAndOutput` parameter defaults to `true`, so AAudio/OpenSLES callers are unaffected.

Apps that opt out accept the OS mic-in-use indicator staying on while muted, in exchange for instant unmute.

## Flow-through for LiveKit consumers

This change is transparent — no client-side API change required. LiveKit's `LocalParticipant.setMicrophoneEnabled(false)` already flows into WebRTC's mute path:

```
LocalParticipant.setMicrophoneEnabled(false)
  → LocalTrackPublication.muted = true              (LocalTrackPublication.kt:44)
  → mediaTrack.enabled = false
  → AudioRtpSender::SetAudioSend(ssrc, enable=false, ...)
  → WebRtcVoiceSendChannel::SetAudioSend            (webrtc_voice_engine.cc:1521)
  → MuteStream(ssrc, muted=true)                    (line 1531)
  → adm->IsStopOnMuteModeEnabled() branch           (line 1736)
       ├─ true (default):  adm->StopRecording()           ← cold start
       └─ false (this PR): adm->SetMicrophoneMute(true)   ← warm mute
```

Apps opt in at ADM construction time:

```kotlin
LiveKit.create(
    appContext = context,
    overrides = LiveKitOverrides(
        audioOptions = AudioOptions(
            javaAudioDeviceModuleCustomizer = { it.setStopRecordingOnMute(false) }
        )
    )
)
```

Existing `setMicrophoneEnabled(...)` call sites keep working. Peers still see the mute indicator, the SFU still drops disabled-track RTP, LiveKit `TrackMuted`/`TrackUnmuted` events fire as before — the only difference is that `AudioRecord` stays warm across the mute cycle.

## Testing

Verified end-to-end on Pixel 6 and USB PTT handset against a `livekit-server --dev` room:

**With `stopRecordingOnMute=false` (this PR's opt-out):**
```
WebRtcVoiceSendChannel::MuteStream: ADM:1
WebRtcAudioRecord: setMicrophoneMute(true)
(AudioRecord stays in STATE_RECORDING; no stopThread / startRecording)
```

**With `stopRecordingOnMute=true` (default, pre-PR behavior):**
```
WebRtcVoiceSendChannel::MuteStream: ADM:1
WebRtcAudioRecord: stopThread
WebRtcAudioRecord: joinAudioRecordThread
(then on unmute: InitRecording + startRecording, audible cold-start delay)
```

Cold-start measured 120–280 ms of dropped speech on Pixel 6, up to 600 ms on a USB gaming headset — vanishes entirely with the opt-out.

**Reproduction:** [maxmini-dev/webrtc-sdk-pr269-test](https://github.com/maxmini-dev/webrtc-sdk-pr269-test) — prebuilt `android_prefixed` AAR at PR HEAD, LiveKit sample app with a runtime toggle for the opt-out, and a step-by-step README (`livekit-server --dev`, `adb install`, log filters).

## Related

- #206 — introduced WebRtcVoiceSendChannel::MuteStream stop-on-mute along with the IsStopOnMuteModeEnabled opt-out that iOS AudioEngineDevice uses; this PR follows the same pattern on Android.

### DIFF
--- a/sdk/android/api/org/webrtc/audio/JavaAudioDeviceModule.java
+++ b/sdk/android/api/org/webrtc/audio/JavaAudioDeviceModule.java
@@ -55,6 +55,7 @@ public class JavaAudioDeviceModule implements AudioDeviceModule {
     private AudioAttributes audioAttributes;
     private boolean useLowLatency;
     private boolean enableVolumeLogger;
+    private boolean stopRecordingOnMute = true;
 
     private Builder(Context context) {
       this.context = context;
@@ -242,6 +243,17 @@ public class JavaAudioDeviceModule implements AudioDeviceModule {
     }
 
     /**
+     * Control whether recording is stopped while all audio tracks are muted. The default is
+     * enabled: muting releases the AudioRecord so the OS mic-in-use indicator turns off. When
+     * disabled, capture keeps running while muted and captured audio is replaced with silence,
+     * making unmute instant at the cost of the mic indicator staying on.
+     */
+    public Builder setStopRecordingOnMute(boolean stopRecordingOnMute) {
+      this.stopRecordingOnMute = stopRecordingOnMute;
+      return this;
+    }
+
+    /**
      * Construct an AudioDeviceModule based on the supplied arguments. The caller takes ownership
      * and is responsible for calling release().
      */
@@ -281,7 +293,8 @@ public class JavaAudioDeviceModule implements AudioDeviceModule {
           new WebRtcAudioTrack(context, audioManager, audioAttributes, audioTrackErrorCallback,
               audioTrackStateCallback, playbackSamplesReadyCallback, useLowLatency, enableVolumeLogger);
       return new JavaAudioDeviceModule(context, audioManager, audioInput, audioOutput,
-          inputSampleRate, outputSampleRate, useStereoInput, useStereoOutput);
+          inputSampleRate, outputSampleRate, useStereoInput, useStereoOutput,
+          stopRecordingOnMute);
     }
   }
 
@@ -440,13 +453,15 @@ public class JavaAudioDeviceModule implements AudioDeviceModule {
   private final int outputSampleRate;
   private final boolean useStereoInput;
   private final boolean useStereoOutput;
+  private final boolean stopRecordingOnMute;
 
   private final Object nativeLock = new Object();
   private long nativeAudioDeviceModule;
 
   private JavaAudioDeviceModule(Context context, AudioManager audioManager,
       WebRtcAudioRecord audioInput, WebRtcAudioTrack audioOutput, int inputSampleRate,
-      int outputSampleRate, boolean useStereoInput, boolean useStereoOutput) {
+      int outputSampleRate, boolean useStereoInput, boolean useStereoOutput,
+      boolean stopRecordingOnMute) {
     this.context = context;
     this.audioManager = audioManager;
     this.audioInput = audioInput;
@@ -455,6 +470,7 @@ public class JavaAudioDeviceModule implements AudioDeviceModule {
     this.outputSampleRate = outputSampleRate;
     this.useStereoInput = useStereoInput;
     this.useStereoOutput = useStereoOutput;
+    this.stopRecordingOnMute = stopRecordingOnMute;
   }
 
   @Override
@@ -463,7 +479,7 @@ public class JavaAudioDeviceModule implements AudioDeviceModule {
       if (nativeAudioDeviceModule == 0) {
         nativeAudioDeviceModule = nativeCreateAudioDeviceModule(context, audioManager, audioInput,
             audioOutput, webrtcEnvRef, inputSampleRate, outputSampleRate, useStereoInput,
-            useStereoOutput);
+            useStereoOutput, stopRecordingOnMute);
       }
       return nativeAudioDeviceModule;
     }
@@ -552,5 +568,5 @@ public class JavaAudioDeviceModule implements AudioDeviceModule {
   private static native long nativeCreateAudioDeviceModule(Context context,
       AudioManager audioManager, WebRtcAudioRecord audioInput, WebRtcAudioTrack audioOutput,
       long webrtcEnvRef, int inputSampleRate, int outputSampleRate, boolean useStereoInput,
-      boolean useStereoOutput);
+      boolean useStereoOutput, boolean stopRecordingOnMute);
 }
--- a/sdk/android/api/org/webrtc/audio/JavaAudioDeviceModule.java
+++ b/sdk/android/api/org/webrtc/audio/JavaAudioDeviceModule.java
@@ -244,9 +244,17 @@ public class JavaAudioDeviceModule implements AudioDeviceModule {
 
     /**
      * Control whether recording is stopped while all audio tracks are muted. The default is
-     * enabled: muting releases the AudioRecord so the OS mic-in-use indicator turns off. When
-     * disabled, capture keeps running while muted and captured audio is replaced with silence,
-     * making unmute instant at the cost of the mic indicator staying on.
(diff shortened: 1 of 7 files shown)


max-buster:
@hiroshihorie @cloudwebrtc — would appreciate a look when you have free time. Restores pre-#206 warm-mute behavior on Android via the same `IsStopOnMuteModeEnabled` opt-out mechanism you added for iOS in #246.

hiroshihorie:
Thanks for picking this up, and for the writeup. The approach is right, and it is what I had in mind when I closed #246. Keeping stop-on-mute as the default and letting platforms that properly implement the ADM mute API opt out individually is exactly it.

A few things I would like to see before this merges. I pushed them as three commits on [`hiroshi/pr269-review-fixes`](https://github.com/webrtc-sdk/webrtc/tree/hiroshi/pr269-review-fixes) so you can cherry-pick whichever ones you agree with.

**[`608d276`](https://github.com/webrtc-sdk/webrtc/commit/608d276396f6c5bd640cc5d54c5b8528e841d8b1)** Make `WebRtcAudioRecord` the source of truth for the mute state. The ADM currently caches it in `microphone_mute_`, but `JavaAudioDeviceModule.setMicrophoneMute()` is public API that writes the same Java field directly, so `MicrophoneMute()` goes stale as soon as an app uses it, and an app level mute gets cleared on the next track unmute. This reads the state back instead of caching, uses the same hook to answer `MicrophoneMuteIsAvailable()` (which still reports the ADM as unable to mute even though `SetMicrophoneMute` now works), and guards the entry points on `initialized_` since `AudioRecordJni` only attaches `env_` in `Init()`.

**[`b766444`](https://github.com/webrtc-sdk/webrtc/commit/b7664440bc76dac395b3e52eef46e5c85e96759b)** Only log on an actual change. `MuteStream` fires on every mute transition, so a warning per call gets noisy once the opt-out is in use.

**[`666b72f`](https://github.com/webrtc-sdk/webrtc/commit/666b72fd9dc64a51220c3bf9420b024a81cbc9e7)** Javadoc. Worth being explicit that recording starts when the track is published and runs until it is unpublished, so the OS mic indicator is lit for the whole published lifetime of the track and not only while unmuted. Also a note that apps should not drive `setMicrophoneMute` themselves in parallel.

One correction for the description: the iOS `IsStopOnMuteModeEnabled` opt-out landed in #206, not #246. #246 was my Linux attempt, which I closed. The guidance you quoted is accurate, just the wrong number.

Nothing here blocks. Threading looks right to me, the checker binds in `Init()` on the worker thread, same as the `StartRecording` and `StopRecording` calls #206 already makes from `MuteStream`. Dropping the unmute restart in the opt-out branch is also fine since `AudioState::AddSendingStream` owns starting recording.

hiroshihorie (approved):
Approving. The three commits above are cleanups rather than blockers, so take whichever ones you want and I will follow up with a separate PR for anything left over.

davidliu (approved):
LGTM, tested on device and both existing default behavior and new feature behavior works.

max-buster:
> Thanks for picking this up, and for the writeup. The approach is right, and it is what I had in mind when I closed #246. Keeping stop-on-mute as the default and letting platforms that properly implement the ADM mute API opt out individually is exactly it.
> 
> A few things I would like to see before this merges. I pushed them as three commits on [`hiroshi/pr269-review-fixes`](https://github.com/webrtc-sdk/webrtc/tree/hiroshi/pr269-review-fixes) so you can cherry-pick whichever ones you agree with.
> 
> **[`608d276`](https://github.com/webrtc-sdk/webrtc/commit/608d276396f6c5bd640cc5d54c5b8528e841d8b1)** Make `WebRtcAudioRecord` the source of truth for the mute state. The ADM currently caches it in `microphone_mute_`, but `JavaAudioDeviceModule.setMicrophoneMute()` is public API that writes the same Java field directly, so `MicrophoneMute()` goes stale as soon as an app uses it, and an app level mute gets cleared on the next track unmute. This reads the state back instead of caching, uses the same hook to answer `MicrophoneMuteIsAvailable()` (which still reports the ADM as unable to mute even though `SetMicrophoneMute` now works), and guards the entry points on `initialized_` since `AudioRecordJni` only attaches `env_` in `Init()`.
> 
> **[`b766444`](https://github.com/webrtc-sdk/webrtc/commit/b7664440bc76dac395b3e52eef46e5c85e96759b)** Only log on an actual change. `MuteStream` fires on every mute transition, so a warning per call gets noisy once the opt-out is in use.
> 
> **[`666b72f`](https://github.com/webrtc-sdk/webrtc/commit/666b72fd9dc64a51220c3bf9420b024a81cbc9e7)** Javadoc. Worth being explicit that recording starts when the track is published and runs until it is unpublished, so the OS mic indicator is lit for the whole published lifetime of the track and not only while unmuted. Also a note that apps should not drive `setMicrophoneMute` themselves in parallel.
> 
> One correction for the description: the iOS `IsStopOnMuteModeEnabled` opt-out landed in #206, not #246. #246 was my Linux attempt, which I closed. The guidance you quoted is accurate, just the wrong number.
> 
> Nothing here blocks. Threading looks right to me, the checker binds in `Init()` on the worker thread, same as the `StartRecording` and `StopRecording` calls #206 already makes from `MuteStream`. Dropping the unmute restart in the opt-out branch is also fine since `AudioState::AddSendingStream` owns starting recording.

[hiroshihorie](https://github.com/hiroshihorie) Thanks for the thorough review and for pushing the fixes as a branch! Cherry-picked all three (081927d, e6c3981, 9f5ee99).

Also fixed the #246 → #206 mixup in the description. Ready on my end whenever you want to merge!

PR #271, opened 2026-07-10
iOS: decouple AudioEngineDevice from RTCAudioSession, surface interruption events

hiroshihorie:
## Problem

`AudioEngineDevice`'s constructor subscribes to `RTCAudioSession`'s delegate to receive interruption and route-change events. That instantiates the `RTCAudioSession` singleton in every app that creates the engine ADM, and the singleton manages the audio session autonomously from its own notification handlers: `updateAudioSessionAfterEvent` calls `setActive:(activationCount > 0)` on interruption end and media-services events. Any audio-session regime in the process that does not keep that activation refcount truthful gets fought by it (a raw-`AVAudioSession` regime reads count 0 and is actively deactivated mid-call). Higher layers that want to own the session lifecycle against plain `AVAudioSession` are forced into refcount bookkeeping purely because the ADM woke the singleton up.

The engine only needs the events, not the session management. Session lifecycle and configuration belong to higher layers, which know about CallKit ownership, background modes, and app audio policy.

## Change

- Observe `NSNotificationCenter` directly with a file-local forwarder feeding the same `AudioSessionObserver` methods, using the same valid-reason filter for route changes that the delegate adapter applied. The `canPlayOrRecord` and output-volume events are dropped: both were log-only no-ops here and are concepts of `RTCAudioSession` itself. Creating the ADM no longer instantiates the singleton; apps that use `RTCAudioSession` directly are unaffected.
- Surface interruption began/ended (with `shouldResume`) through `AudioDeviceObserver` and new optional `RTCAudioDeviceModuleDelegate` methods. With the singleton's implicit count-gated recovery out of the picture, a session-owning layer needs the signal to reactivate on resume. The events fire on the device thread before the engine reacts (in particular before the engine restarts on interruption end), matching the guidance to activate the session before restarting audio I/O. They are fire-and-forget and must not block, unlike the engine lifecycle hooks. Optional and `respondsToSelector`-guarded, so existing delegates keep working unchanged.

## Compatibility

- The legacy `AudioDeviceIOS` path still uses `RTCAudioSession` and is untouched.
- Existing delegates (LiveKit Swift SDK, React Native) compile and behave as before; adopting the new interruption events is opt-in.
- No session `setActive:` or configuration calls are added to the ADM.

## Testing

Compile-verified for iOS device (arm64, debug) per object and as the full `framework_objc` build. Runtime verification of interruption flows on device is pending; downstream SDK adoption (moving the LiveKit session observers to plain `AVAudioSession` and the new events) will follow in the SDK repos.

Related: #270 (enable-rollback compensation, standalone) completes the delegate contract for session-owning observers.

### DIFF
--- a/api/audio/audio_device.h
+++ b/api/audio/audio_device.h
@@ -314,6 +314,14 @@ class AudioDeviceObserver {
 
   virtual int32_t OnEngineWillRelease(AVAudioEngine* engine) { return 0; }
 
+  // Audio session interruption events (iOS). Fired on the device thread before
+  // the engine reacts to the event, so an observer that owns the audio session
+  // lifecycle can act first. In particular OnAudioSessionInterruptionEnded fires
+  // before the engine restarts, giving the owner a chance to reactivate the
+  // session. Implementations must not block.
+  virtual void OnAudioSessionInterruptionBegan() {}
+  virtual void OnAudioSessionInterruptionEnded(bool should_resume) {}
+
   // Override the input node configuration with a custom implementation.
   virtual int32_t OnEngineWillConnectInput(AVAudioEngine* engine, AVAudioNode* src,
                                            AVAudioNode* dst, AVAudioFormat* format,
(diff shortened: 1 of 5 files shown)


pblazej:
Out of curiosity: do we have any counter-examples (issues) here? Similar to https://github.com/livekit/client-sdk-swift/issues/886

PR #236, opened 2026-05-11
Add getter for track volume

hiroshihorie:
* Fixed: ObjC reporting 0.0 when setter is never called
 * Expose for Android also

PR #273, opened 2026-07-14
feat(m150 upgrade): patch from m144_release (#173) (#217)

cloudwebrtc:


pblazej:
Reviewed locally on macOS arm64.

**Method.** Rather than re-reading the full 418-file diff, I compared the fork's patch at m144 (`8f3537ef5b…m144_release`) against the fork's patch at m150 (`1f975dfd76…<PR head>`). **289 of the 418 fork-touched files are byte-identical replays** — clean. The 129 files that had to be adapted are what I reviewed.

**Tests.** `rtc_unittests`: **2345/2347 pass**, 2 skipped (`DenormalDisabler.*`, platform-dependent), ~50s. Two build fixes were needed first (B1, B3 below).

---

## Blockers

### B1 — Committed merge conflict markers

`pc/rtp_sender_receiver_unittest.cc:799` and `:814`:

```
../../pc/rtp_sender_receiver_unittest.cc:799:1: error: version control conflict marker in file
../../pc/rtp_sender_receiver_unittest.cc:814:1: error: version control conflict marker in file
```

`run_loop_.Flush()` is the m144 API; m150 renamed it to `FlushWorker()`. Both sides should be kept — the fork's `GetVolume()` assertion from #236 plus the m150 flush:

```cpp
  audio_track_->GetSource()->SetVolume(0.5);
  EXPECT_EQ(0.5, audio_track_->GetSource()->GetVolume());
  // Wait for the worker thread to apply the volume change.
  FlushWorker();
```

and the same shape for the `0.8` case below it. Verified locally: compiles clean and `RtpSenderReceiverTest.RemoteAudioTrackSetVolume` passes.

### B2 — #174 regressed (missing `RTC_OBJC_TYPE`)

`sdk/objc/api/peerconnection/RTCTracing.mm:17` defines `void RTCSetupInternalTracer(void)`, but `RTCTracing.h:15` still declares `RTC_EXTERN void RTC_OBJC_TYPE(RTCSetupInternalTracer)(void)`.

Under `rtc_objc_prefix=LK` the declared symbol is `LKRTCSetupInternalTracer` while the definition is `RTCSetupInternalTracer` → undefined symbol for every downstream caller. There are no in-tree callers, so no build in this repo catches it; it surfaces in client-sdk-swift. The `Environment` plumbing added on the same line is correct — only the macro was dropped.

### B3 — `rtc_unittests` does not link on macOS

`modules/audio_device/BUILD.gn` compiles `audio_engine_device.mm` for mac (line 363), but `audio_device_impl_frameworks` (line 420) lists only ApplicationServices / AudioToolbox / CoreAudio / CoreGraphics — no `AVFoundation.framework`:

```
Undefined symbols for architecture arm64:
  "_AVAudioEngineConfigurationChangeNotification", referenced from: ...
  "_AVMediaTypeAudio", referenced from: ...
  "_OBJC_CLASS_$_AVAudioEngine", "_OBJC_CLASS_$_AVCaptureDevice", ...
```

**Pre-existing — m144 has the identical gap**, so not a regression from this port, but it blocks any test build. One-line fix.

---

## Functional regressions from the port

### R1 — Default-device GUID is now empty

`modules/audio_device/mac/audio_device_mac.cc:1684`.

m144 did `SimpleStringBuilder guid_ss(guid); guid_ss << "default";` before the branch, so default devices reported `guid == "default"`. The rewrite moved the guid write *inside* the `else` branch (`:1705-1708`); the `isDefaultDevice` branch now writes only `name`. Since `PlayoutDeviceName`/`RecordingDeviceName` memset guid to 0 first, default devices report an empty GUID.

Minor asymmetry in the same function: the guid path guards `guid.size() > 0`, the name path does `name.size() - 1` unguarded. Unreachable today (all callers pass `kAdmMaxDeviceNameSize`), but worth matching.

### R2 — Dead jitter-buffer members

`media/engine/webrtc_voice_engine.h:181-185` re-adds `audio_jitter_buffer_max_packets_`, `_fast_accelerate_` and `_min_delay_ms_`, which upstream M150 deleted. They are written in `ApplyOptions` (`.cc:633-644`) and **never read**.

M150 routes these through the channel-level `audio_config_` (`.cc:2238-2257`, consumed at `:2427`); on m144 the engine-level fields fed recv-stream creation directly (`m144 .cc:2382-2384`). As committed, engine/factory-level jitter options set before any channel exists are silently dropped. Either delete the members or wire them into `audio_config_`.

---

## Downstream-visible API changes

### A1 — Required ObjC protocol method changed

`sdk/objc/api/peerconnection/RTCAudioDeviceModule.h:147` — `audioDeviceModule:willEnableEngine:isPlayoutEnabled:isRecordingEnabled:` gained a required `isVoiceProcessingEnabled:` parameter. Every implementer must update in lockstep.

Worth noting: `m144_release` still carries the 4-arg form despite also having #275, so the two branches now expose incompatible protocols.

### A2 — Silent no-op property

`sdk/objc/components/audio/RTCAudioProcessingConfig.mm:26` — `echoCancellerMobileMode` is now backed by a standalone `BOOL` since M150 removed `EchoCanceller::mobile_mode`. Apps setting it get no effect and no diagnostic. Consider `__attribute__((deprecated("no-op since M150")))` on the property so this fails at compile time rather than in the field.

hiroshihorie (approved):

PR #277, opened 2026-08-03
Fall back to the OS trust store for TURN/TLS

pblazej:
## Problem

TURN/TLS trust anchors come from `rtc_base/ssl_roots.h`, generated from `https://pki.goog/roots.pem` — Google Trust Services' own chain of trust, not a general CA store. It holds 36 roots from 2023-05-09, one already expired (Baltimore CyberTrust, May 2025), and is missing Amazon Root CA 1–4, Starfield Services Root CA - G2 and ISRG Root X1/X2.

So a TURN server behind an AWS NLB with an ACM certificate, or anything issued by Let's Encrypt, fails its handshake with `unknown_ca` — even though the server sends a complete chain and the host OS trusts it. Reported downstream as livekit/client-sdk-swift#1074 and livekit/rust-sdks#1301.

Worth being precise: nothing is missing from the wire. `s3.amazonaws.com` sends leaf → `Amazon RSA 2048 M04` → `Amazon Root CA 1` (cross-signed by Starfield Services G2). It fails purely on anchors.

## Approach

Where the platform exposes a trust evaluation API, use it as a fallback when the embedder hasn't supplied its own verifier. `SSLVerifyInternal` only consults a verifier *after* the built-in anchors have failed to produce a path, so this can widen what's accepted and never narrow it — no currently-working connection changes behaviour.

Two details worth calling out:

- **The whole chain is passed.** The `SSLCertificateVerifier` exposed via the ObjC and JNI bridges forwards only `chain.Get(0)`. With a leaf alone the OS must recover intermediates over AIA — a plaintext HTTP fetch to the CA, on the network thread, mid-handshake. Measured ~125 ms cold, and it fails outright where port 80 is blocked or where the certificate carries no AIA extension (common for private CAs). `SecTrustSetNetworkFetchAllowed(false)` is set explicitly since nothing needs fetching.
- **Hostname matching is untouched.** `SSLPostConnectionCheck` already calls `VerifyPeerCertMatchesHost` independently of the verification result, so no name is passed to the platform policy.

## Platform coverage

| platform | status |
|---|---|
| Apple (macOS/iOS/tvOS/visionOS/Catalyst) | implemented, one file |
| Windows | implemented, **not compiled** — no toolchain available here |
| Linux / BSD | no OS API exists; keeps built-in anchors |
| Android | has an API, but `X509TrustManager` needs a JNI target outside `rtc_base` — follow-up |

## Verification

- `ninja rtc_base:ssl_adapter` and `sdk:mac_framework_bundle` build clean on macOS arm64.
- The built framework imports `SecTrustCreateWithCertificates`, `SecTrustEvaluateWithError`, `SecTrustSetNetworkFetchAllowed`, `SecPolicyCreateSSL`, `SecCertificateCreateWithData`.
- SecTrust behaviour validated separately against real chains: full chain verifies in ~1.5 ms with network fetching off; leaf-only fails without AIA.

**Draft because** the Windows path is uncompiled, Android is unimplemented, and there's no end-to-end relay test in this branch yet — a positive one needs a TURN server whose chain is trusted by the OS but absent from the 36.

## Note on the alternative

This does not remove the need to regenerate `ssl_roots.h`. Linux and Android still depend on it, and today's `pki.goog` has shrunk to 19 roots, so refreshing from the same source would make things worse. Regenerating from `https://curl.se/ca/cacert.pem` (119 roots, named in `generate_sslroots.py --help`) is a one-file change that fixes every platform at once, at +91 KiB per arch slice. The two are complementary; that one is arguably the cheaper first move.

### DIFF
--- a/rtc_base/platform_certificate_verifier_apple.cc
+++ b/rtc_base/platform_certificate_verifier_apple.cc
@@ -125,7 +125,8 @@ class AppleCertificateVerifier final : public SSLCertificateVerifier {
 
 }  // namespace
 
-std::unique_ptr<SSLCertificateVerifier> CreatePlatformCertificateVerifier() {
+std::unique_ptr<SSLCertificateVerifier>
+CreateNativePlatformCertificateVerifier() {
   return std::make_unique<AppleCertificateVerifier>();
 }
 
--- a/rtc_base/platform_certificate_verifier_apple.cc
+++ b/rtc_base/platform_certificate_verifier_apple.cc
@@ -60,6 +60,10 @@ class AppleCertificateVerifier final : public SSLCertificateVerifier {
       return false;
     }
 
+    // Appended in the order received. SSLCertificateVerifier::VerifyChain is
+    // documented to deliver the chain leaf first, then intermediates, which is
+    // also what SecTrustCreateWithCertificates expects: it evaluates element 0
+    // and treats the remainder as candidate intermediates.
     for (size_t i = 0; i < chain.GetSize(); ++i) {
       Buffer der;
       chain.Get(i).ToDER(&der);
--- a/rtc_base/platform_certificate_verifier_apple.cc
+++ b/rtc_base/platform_certificate_verifier_apple.cc
@@ -12,8 +12,6 @@
 #include <Security/Security.h>
 
 #include <memory>
-#include <utility>
-#include <vector>
 
 #include "rtc_base/buffer.h"
 #include "rtc_base/logging.h"
@@ -47,9 +45,7 @@ class ScopedCF {
 class AppleCertificateVerifier final : public SSLCertificateVerifier {
  public:
   bool Verify(const SSLCertificate& certificate) override {
-    std::vector<std::unique_ptr<SSLCertificate>> single;
-    single.push_back(certificate.Clone());
-    return VerifyChain(SSLCertChain(std::move(single)));
+    return VerifyChain(SSLCertChain(certificate.Clone()));
   }
 
   bool VerifyChain(const SSLCertChain& chain) override {
@@ -100,15 +96,28 @@ class AppleCertificateVerifier final : public SSLCertificateVerifier {
 
     // The peer sent the whole chain, so nothing needs to be fetched. Leaving
     // this enabled would allow an AIA lookup to block the network thread for
-    // the duration of a plaintext HTTP round trip mid-handshake.
+    // the duration of a plaintext HTTP round trip mid-handshake. Note that this
+    // also stops OCSP and CRL fetching, so revocation is decided on whatever
+    // the system has already cached; a blocking revocation fetch on this thread
+    // is the worse of the two.
     SecTrustSetNetworkFetchAllowed(trust.get(), false);
 
     CFErrorRef error = nullptr;
     const bool trusted = SecTrustEvaluateWithError(trust.get(), &error);
     ScopedCF<CFErrorRef> scoped_error(error);
     if (!trusted) {
-      RTC_LOG(LS_INFO) << "Peer certificate chain was rejected by the system "
-                          "trust store.";
+      ScopedCF<CFStringRef> reason(
+          error != nullptr ? CFErrorCopyDescription(error) : nullptr);
+      char buffer[256] = {};
+      if (reason && CFStringGetCString(reason.get(), buffer, sizeof(buffer),
+                                       kCFStringEncodingUTF8)) {
+        RTC_LOG(LS_INFO) << "Peer certificate chain was rejected by the system "
+                            "trust store: "
+                         << buffer;
+      } else {
+        RTC_LOG(LS_INFO) << "Peer certificate chain was rejected by the system "
+                            "trust store.";
+      }
     }
     return trusted;
   }
--- a/rtc_base/platform_certificate_verifier_apple.cc
+++ b/rtc_base/platform_certificate_verifier_apple.cc
@@ -1,0 +1,123 @@
+/*
+ *  Copyright 2026 The WebRTC Project Authors. All rights reserved.
+ *
+ *  Use of this source code is governed by a BSD-style license
+ *  that can be found in the LICENSE file in the root of the source
+ *  tree. An additional intellectual property rights grant can be found
+ *  in the file PATENTS.  All contributing project authors may
+ *  be found in the AUTHORS file in the root of the source tree.
+ */
+
+#include <CoreFoundation/CoreFoundation.h>
+#include <Security/Security.h>
+
+#include <memory>
+#include <utility>
+#include <vector>
+
+#include "rtc_base/buffer.h"
+#include "rtc_base/logging.h"
+#include "rtc_base/platform_certificate_verifier.h"
+#include "rtc_base/ssl_certificate.h"
+
+namespace webrtc {
+namespace {
+
+template <typename T>
+class ScopedCF {
+ public:
+  ScopedCF() = default;
+  explicit ScopedCF(T ref) : ref_(ref) {}
+  ScopedCF(const ScopedCF&) = delete;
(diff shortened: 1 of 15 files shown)


pblazej:
## Verification

**macOS arm64** — built, unit tested, and exercised end to end.

`rtc_base_unittests` additions, 5/5 passing, nothing skipped (zero skips confirms the Apple verifier is present and actually run, not compiled out):

```
[  PASSED  ] 5 tests.   SkippedTestCount= 0
```

End to end against local coturn, using a relay-only `LKRTCPeerConnection` on a locally built `WebRTC.framework`:

```
(openssl_adapter.cc:919): Invoking SSL Verify Callback.
(platform_certificate_verifier_apple.cc:110): Peer certificate chain was
    rejected by the system trust store.
RESULT relay_candidates=0 ice_errors=2
```

Those lines are the wiring proof the unit tests cannot give: the verifier is installed as the default with no application involvement, it is reached only after the built-in anchors have already failed, and it receives the chain and evaluates it. A control run over plaintext `turn:` on the same server yields `relay_candidates=1`, so certificate verification is the only variable.

**Not verified.** The accept path — a chain the OS trusts producing a relay candidate — needs a CA in the machine's trust store, so it is not covered here. Windows is written but never compiled. Android is written but never built or run: `build/config/BUILDCONFIG.gn:244` asserts `host_os == "linux"` and the NDK ships only `linux-x86_64` prebuilts, so neither a cross build nor an emulator is possible from macOS. Both need a Linux builder before this leaves draft.

## Reproducing on any SDK

Steps 1, 2 and 6 need no SDK build at all, which is why this reproduced identically on Swift and the Python FFI.

1. Extract the anchor set: dump `rtc_base/ssl_roots.h` to PEM. The same 36 roots ship on every platform.
2. Pick a TURN server outside it — anything AWS ACM (`-> Amazon Root CA 1`, cross-signed by Starfield Services G2) or Let's Encrypt (`-> ISRG Root X1`):
   ```
   openssl s_client -connect <turn-host>:443 -CAfile webrtc_roots.pem   # error 20 at depth 2
   openssl s_client -connect <turn-host>:443                            # code 0
   ```
3. Force relay-only so no direct path masks it: `.relay` (Apple), `RTCConfiguration.iceTransportsType = RELAY` (Android), `ice_transport_type = TRANSPORT_RELAY` (Rust/Python).
4. Read the native logs — signaling and the join response succeed, then ICE 701 with no relay candidate. Look for `openssl_adapter.cc` errors rather than TURN allocation errors to tell TLS from reachability. `adb logcat -s libjingle:*` after `Logging.enableLogToDebugOutput(Severity.LS_INFO)`; `RTCSetMinDebugLogLevel(.info)` on Apple.
5. Control: same server and credentials over `turn:` instead of `turns:`. Relay candidates appear.
6. Confirm at the wire: `tcpdump` on 443 shows ServerHello and the certificate, then a client TLS fatal alert 48 `unknown_ca`, and no TURN Allocate.

Note for anyone picking a test server: LiveKit Cloud will not reproduce it. Its chain is `*.turn.livekit.cloud -> ZeroSSL -> Sectigo R46 -> USERTrust RSA`, and USERTrust RSA is one of the 36, so it verifies against the stock bundle. This is a self-hosted-only failure.

pblazej:
@davidliu @cloudwebrtc I verified 🍏 e2e, can you try confirming Win/Android behavior?

pblazej:
Build: https://github.com/webrtc-sdk/webrtc-build/actions/runs/30811347947 🔴 
Build: https://github.com/webrtc-sdk/webrtc-build/actions/runs/30813447488

davidliu:
@pblazej will check it out tomorrow. Just to double check, the android code looks self contained and there's nothing that needs configuring client side right?

pblazej:
> there's nothing that needs configuring client side right

Yes it should be transparent to the SDK.

Also @davidliu alternative injection method: ff73474593ec2a6127c352fcad7bf798bff30250 that's beyond my jni knowledge, feel free to revert it.

davidliu (approved):
LGTM. Verified works on Android (also discovered a bug with custom RTCConfigurations messing with using relay servers in the Android SDK). The new injection method in JNI_OnLoad looks fine as well.

pblazej:
@cloudwebrtc I'll wait for your windows build then 😸

cloudwebrtc:
Just committed a fix for the win build, and the e2e tests passed.
```ps1
rtc_unittests.exe --gtest_filter=PlatformCertificateVerifierTest.*
```

image

cloudwebrtc (approved):

rtc_base/platform_certificate_verifier_apple.cc
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    ScopedCF<CFMutableArrayRef> certs(
+        CFArrayCreateMutable(kCFAllocatorDefault, chain.GetSize(),
+                             &kCFTypeArrayCallBacks));
+    if (!certs) {
+      return false;
+    }
+
+    for (size_t i = 0; i < chain.GetSize(); ++i) {
```
xianshijing-lk, line 63: A certificate chain usually has a direction:

Root CA ->  Intermediate CA -> Leaf certificate (the server certificate)

I believe the |chain| should be in the right order, but might not be guaranteed ?
and wonder if we should add a check like 
RTC_DCHECK(chain.Get(0).IsLeaf());

rtc_base/platform_certificate_verifier_win.cc
```diff
@@ -0,0 +1,145 @@
...
+                      CERT_STORE_CREATE_NEW_FLAG | CERT_STORE_DEFER_CLOSE_UNTIL_LAST_FREE_FLAG,
+                      NULL);
+    if (!intermediates) {
+      return false;
+    }
+
+    for (size_t i = 1; i < chain.GetSize(); ++i) {
+      Buffer der;
+      chain.Get(i).ToDER(&der);
+      CertAddEncodedCertificateToStore(
```
xianshijing-lk, line 79: I think CertAddEncodedCertificateToStore returns a bool according to https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-certaddencodedcertificatetostore 

should we handle the failure ?

rtc_base/platform_certificate_verifier_win.cc
```diff
@@ -0,0 +1,145 @@
...
+          intermediates, X509_ASN_ENCODING | PKCS_7_ASN_ENCODING, der.data(),
+          static_cast<DWORD>(der.size()), CERT_STORE_ADD_ALWAYS, NULL);
+    }
+
+    ScopedCertContext leaf(CreateContext(chain.Get(0)));
```
xianshijing-lk, line 84: maybe put a comment to explain why chian.Get(0) is expected to be the leaf ?

xianshijing-lk (approved):
some bits, lgtm

rtc_base/platform_certificate_verifier_win.cc
pblazej: Done in 63e19565f0.

rtc_base/platform_certificate_verifier_win.cc
pblazej: Done in 63e19565f0.

rtc_base/platform_certificate_verifier_apple.cc
pblazej: Ordering is guaranteed by contract:

```
rtc_base/ssl_certificate.h:135
  // Verify a complete certificate chain (leaf first, then intermediates).
```

Both producers in `OpenSSLAdapter` preserve it — neither reorders. I've added a comment saying so in 63e19565f0, since both implementations rely on element 0 being the leaf.

On the DCHECK: `IsLeaf()` doesn't exist, and `SSLCertificate` has no basic constraints accessor to build it from.

PR #283, opened 2026-08-20
feat: expose enableSctpSnap on RTCConfiguration (ObjC and Android) for WARP

hiroshihorie:
## Description

Exposes the native `RTCConfiguration.enable_sctp_snap` field (m150, [draft-hancke-tsvwg-snap](https://www.ietf.org/archive/id/draft-hancke-tsvwg-snap-00.html)) through the ObjC and Android SDK wrappers so client SDKs can enable SNAP (SCTP INIT in SDP), the data channel half of WARP.

- ObjC: `enableSctpSnap` property on `RTCConfiguration`, mapped in both directions (init from native defaults, `createNativeConfiguration`)
- Android: `enableSctpSnap` field on `PeerConnection.RTCConfiguration` with a `CalledByNative` getter, mapped in the JNI conversion

Defaults to false on both platforms, so this is a no-op for existing users.

## Why the config field instead of the `WebRTC-Sctp-Snap` field trial

`enable_sctp_snap` is not in the modifiable list checked by `ApplyConfiguration`. If the field trial flips it to true inside the peer connection while the wrapper keeps passing false, every later `setConfiguration` call fails with INVALID_MODIFICATION. Carrying it on the configuration object keeps creation and reconfiguration consistent. This mirrors how livekit/rust-sdks#1342 handled it.

## SPED (the other half of WARP)

No new API needed. The DTLS-in-STUN field trial `WebRTC-IceHandshakeDtls` is already reachable via `+[RTCPeerConnectionFactory configureFieldTrials:]` on ObjC and `InitializationOptions.setFieldTrials` on Android.

### DIFF
--- a/sdk/android/api/org/webrtc/PeerConnection.java
+++ b/sdk/android/api/org/webrtc/PeerConnection.java
@@ -586,6 +586,16 @@ public class PeerConnection {
      */
     public boolean enableIceGatheringOnAnyAddressPorts;
 
+    /**
+     * Enable SNAP (SCTP INIT in SDP), part of WARP. Saves a round trip on
+     * data channel setup when the remote peer also supports it.
+     * See: https://www.ietf.org/archive/id/draft-hancke-tsvwg-snap-00.html
+     * This maps to an immutable field of the native configuration, so it must
+     * hold the same value at peer connection creation and on every
+     * setConfiguration call.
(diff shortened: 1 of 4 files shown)


cloudwebrtc (approved):
lgtm!

PR #284, opened 2026-08-21
Guard simulcast-factory AV1 advertisement; fail fast on nil ObjC encoder/decoder

pblazej:
`RTCVideoEncoderFactorySimulcast` advertised AV1 unconditionally, so builds with `enable_libaom=false` (e.g. the stripped Apple variant, webrtc-sdk/webrtc-build#64) offer AV1 in SDP while `createEncoder:` cannot create it — the nil encoder gets wrapped and nil-messaging makes `InitEncode` report success, yielding a silently dead track. This guards the AV1 entry with `RTC_USE_LIBAOM_AV1_ENCODER`, matching `RTCDefaultVideoEncoderFactory`. As defense in depth, the ObjC native encoder/decoder factories now return `nullptr` with an error log when the wrapped factory returns nil, instead of wrapping nil. No behavior change for regular builds where AV1 is compiled in. The same change was compile- and link-verified on m144 in both regular and stripped configurations; m150 build verification of this branch is in progress and will be confirmed in a comment.

### DIFF
--- a/sdk/objc/components/video_codec/RTCVideoEncoderFactorySimulcast.mm
+++ b/sdk/objc/components/video_codec/RTCVideoEncoderFactorySimulcast.mm
@@ -53,11 +53,13 @@
         [addingCodecs addObject: codec];
     }
 
+#if defined(RTC_USE_LIBAOM_AV1_ENCODER)
     auto av1Format = webrtc::SdpVideoFormat(
          webrtc::kAv1CodecName, webrtc::CodecParameterMap(),
          webrtc::LibaomAv1EncoderSupportedScalabilityModes());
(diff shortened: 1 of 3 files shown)


pblazej:
m150 verification done: built this branch (ccf075c85a) locally as iOS arm64 device release in both the regular config and the stripped config (`enable_libaom=false rtc_include_dav1d_in_internal_decoder_factory=false rtc_libvpx_build_vp9=false optimize_for_size=true`). Both compile and link. The stripped binary contains zero libaom/dav1d traces and retains all 66 H265 exports; the regular binary is unchanged (the guard is a no-op when AV1 is compiled in).

hiroshihorie (approved):

